PR #10, opened 2015-07-20
[Fix] Interface TelegramHttpClientInterface not found

codengine:
The bug prevented the construction of Telegram objects with a custom HTTP handler, as the Interface could not be found.


irazasyed:
@codengine Thanks for the PR. Merged.

PR #17, opened 2015-08-07
Loading file like resource

rkhitin:
Now file can be load like resource. For example:   `$telegram->sendDocument($chatId, fopen($imageUrl));`

### DIFF
--- a/src/Telegram.php
+++ b/src/Telegram.php
@@ -574,10 +574,12 @@ class Telegram
                 continue;
             }
 
-            if (is_file($contents)) {
-                $contents = (new InputFile($contents))->open();
-            } else {
-                $contents = (string) $contents;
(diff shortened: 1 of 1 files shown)


jonnywilliamson:
What benefit does this give us?

Does it allow us to send a file that's not local? Ie 

`$telegram->sendDocument($chatId, fopen('http://example.com/demo.pdf'))`

If so, then +1

rkhitin:
Yes. That was the goal.

irazasyed:
Have you tested this? @rkhitin

rkhitin:
Of course, it's working in real project right now. And, actually, there only two lines was changed.

irazasyed:
Thanks for the PR. Merged. Will tag the update soon.

PR #19, opened 2015-08-19
Added new features from August API Changelog

jonnywilliamson:
As from the latest [Official API changelog](https://core.telegram.org/bots/api-changelog#august-15-2015)

>August 15, 2015
>Added new type Voice and new method sendVoice for sending voice messages.
>Earlier Audio and sendAudio should now be used for sending music files. Telegram clients will show such files in the in-app music player. If you were using sendAudio for your bot to send voice messages, please use sendVoice instead.
>Added optional fields performer, title to the Audio object and sendAudio method.
>Added optional field voice to the Message object.

This PR adds all new features in the above changelog to the the bot sdk.

### DIFF
--- a/src/Objects/Message.php
+++ b/src/Objects/Message.php
@@ -19,6 +19,7 @@ class Message extends BaseObject
             'photo' => PhotoSize::class,
             'sticker' => Sticker::class,
             'video' => Video::class,
+            'voice' => Voice::class,
             'contact' => Contact::class,
             'location' => Location::class,
             'new_chat_participant' => User::class,
--- a/src/Objects/Voice.php
+++ b/src/Objects/Voice.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Irazasyed\Telegram\Objects;
+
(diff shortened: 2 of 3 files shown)


irazasyed:
Thanks @jonnywilliamson for the PR. This is a breaking change though. Will tag soon after i verify a few more things that seems to have been changed.

jonnywilliamson:
@irazasyed What other things did I miss that had changed ? Just curious!

irazasyed:
I would want to keep the param order as per the official docs and that's what is being followed in all other functions mostly. Because if we change the order, then the users would have to start going through this library's own documentation and end up confusing or learning twice which i feel is unnecessary and the whole point of this library was to make things easier, not the other way around. 

As far as other things are concerned, I thought there are other things we missed but didn't find any (Already implemented actually). So left that for now. Since this is a major change and would require me to bump to `v1.0.0` i was thinking to do that with the command handler system, So it would make sense we have a new feature with some major changes, Would give a good reason for people to upgrade :)

PR #37, opened 2015-09-18
More message detection

jonnywilliamson:
See below

### DIFF
--- a/src/Api.php
+++ b/src/Api.php
@@ -665,33 +665,40 @@ class Api
     /**
      * Determine if a given type is the message.
      *
-     * @param string $type
-     * @param Update $update
+     * @param string         $type
+     * @param Update|Message $object
      *
      * @return bool
      */
-    public function isMessageType($type, Update $update)
+    public function isMessageType($type, $object)
     {
-        if ($update->getMessage()->has(strtolower($type))) {
+        if ($object instanceof Update) {
+            $object = $object->getMessage();
+        }
+
+        if ($object->has(strtolower($type))) {
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


jonnywilliamson:
OK, so this PR is just a very small tweak but I thought I'd explain why I needed/wanted it.

We have two functions in API. 

isMessageType()
detectMessageType()

Both currently require an UPDATE object to see what type of message we have.



However, I've found a use case in which I actually needed to detect the message type from a MESSAGE object.

Basically, when you send a message (eg a photo) with Telegram::sendPhoto you always get a MESSAGE object as the return.

In my program, I fire an event after I send a message, and I pass the MESSAGE object through to the event. 

Now when my listener listens for a "sent message event", it receives the MESSAGE object. At this stage, I do different things depending on the type of message that was sent. To work that out, I wanted to use the detectMessageType() function, only I couldn't because it needed an UPDATE object instead.

So, I tweaked the two functions to allow them to accept EITHER object and return the correct reply.

Of course I could create many events like photoMessageSent, videoMessageSent etc etc and listen for each one, but that was going to add a lot more code, rather than listen for a generic messageSent event and do a quick detection afterwards.


Anyway, I hope that makes sense. It's only a few lines of code changed, and I have tested and it works well with bothe UPDATE and MESSAGE objects so no BC problems.


By the way, I'm still not sure how to do a PR that doesn't make you have to do a merge commit. I think you said something about creating a branch and doing it from there. So that's what I've done. If that's wrong, don't merge it - tell me what to do and I'll do it properly. I would like to learn the correct way.

PR #40, opened 2015-09-22
Use POST requests when contacting the telegram api server.

jonnywilliamson:
Whilst a GET request according to the specs may have an unlimited number of characters, in practice it appears that different versions of PHP and even webbrowsers do have a limit.

http://stackoverflow.com/questions/3091485/what-is-the-limit-on-querystring-get-url-parameters
http://stackoverflow.com/questions/812925/what-is-the-maximum-possible-length-of-a-query-string
http://stackoverflow.com/questions/7724270/max-size-of-url-parameters-in-get

This means that if we have a very long message etc and try to use a GET request, the request will likely be corrupt.

We can bypass this limitation by using a POST request.

Telegram supports four ways of passing parameters, but the most useful are:

`application/x-www-form-urlencoded`
`multipart/form-data` (use to upload files)

This PR, switches the normal requests to use POST instead of GET and ensure the parameters are passed via `application/x-www-form-urlencoded`

If a file is being uploaded, it automatically knows to use `multipart/form-data`.

No breaking changes.

### DIFF
--- a/src/Api.php
+++ b/src/Api.php
@@ -6,10 +6,10 @@ use Telegram\Bot\Objects\File;
 use Telegram\Bot\Objects\User;
 use Telegram\Bot\Objects\Update;
 use Telegram\Bot\Objects\Message;
-use Telegram\Bot\Objects\UserProfilePhotos;
-use Telegram\Bot\FileUpload\InputFile;
 use Telegram\Bot\Commands\CommandBus;
+use Telegram\Bot\FileUpload\InputFile;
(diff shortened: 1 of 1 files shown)


irazasyed:
Thank you @jonnywilliamson :)

jonnywilliamson:
@irazasyed No probs!

Any chance of a tag release?

irazasyed:
Will do that sometime this week. Gotta do some testing of all the recent changes.

Issue #22, opened 2015-09-01
[Todo] Add Multi-Bot & Multi/Shared Commands Support!


irazasyed:
Would a multi-bot support be good? Looking for feedback.

I have plans to add but not on my priority list unless there's a requirement.

jonnywilliamson:
Just so I'm clear in my mind,

Basically just allow you to control more than 1 bot from the same code base? ie Multiple api keys in config files etc?

Personally I don't have a need for this *now*, but I could see how that would be helpful to people.

irazasyed:
Exactly!

Multiple API Keys, Same Code Base and Multiple/Shared Commands using the commands handler system.

This will come handy like in case where lets say there are 2 different bots in a group chat (Your own), Based on whichever bot the user sends a command to, the library will automatically trigger its appropriate command and respond accordingly. 

Will also help in maintaining one setup to handle all your bots, Should make things easier instead of handling multiple setups for bot developers. I personally don't have multiple bots for now but i do have plans to create a few, maybe at that time i would feel a need for this.

Ham3D:
+1

irazasyed:
Just pushed multi-bot support: https://github.com/irazasyed/telegram-bot-sdk/commit/e74f71492fc92d584be473e256a2b5d91a69b8ed

Currently it works only with Laravel but obviously will add support for standalone users too.

Your routes could look like this (Basic example):
```php
Route::get('/{bot}/set-webhook', function ($bot = 'common') {
    $response = Telegram::bot($bot)->setWebhook(['url' => 'https://domain.com/'.$bot.'/webhook']);
});

Route::post('/{bot}/webhook',  function ($bot = 'common') {
    $update = Telegram::bot($bot)->commandsHandler(true);

    return 'Thank You Telegram!';
});
```

If you guys are interested, feel free to test it using the master branch / `3.0-dev` and leave your feedback. Also make sure to refer the config file as it has all the explanation and examples of how the new system works.

irazasyed:
P.S I'm yet to update the commands handler to handle commands in a telegram group for multiple bots. Feel free to send a PR if you work on it though.

MostafaQanbaryan:
Sorry to open this up, But does this works for Standalone now?

irazasyed:
@MostafaQanbaryan No problem. No, Standalone is not yet supported. It'll be supported before it's tagged.

PR #43, opened 2015-10-02
Add list of bots allowed to use a command

### LINKED ISSUE #33: command detection
Imagine my username's bot is "something_bot".

when a command receives like that:
/start 
StartCommand triggered.

and when this command receives
/start@something_bot
again StartCommand triggered.

and bug is here:
when
/start@ANOTHER_bot
receives StartCommand triggered, the command is not for my bot.

parshikov:
Add an extra array in Command class to specify the bot usernames which can execute this command.
If this array is not set, this command is available to run. Fix #33

### DIFF
--- a/src/Api.php
+++ b/src/Api.php
@@ -58,6 +58,11 @@ class Api
     protected $commandBus = null;
 
     /**
+     * @var string|null Telegram bot's username.
+     */
+    protected $name = null;
+
+    /**
      * Instantiates a new Telegram super-class object.
      *
      *
@@ -228,6 +233,21 @@ class Api
         $response = $this->post('getMe');
 
         return new User($response->getDecodedBody());
+    }
+
+    /**
+     * A simple method for getting your bot's name.
+     * Returns string information about the bot's name.
+     *
+     * @return string
+     */
+    public function getName()
+    {
+        if (!isset($this->name))
+        {
+            $this->name = $this->getMe()['username'];
+        }
+        return $this->name;
     }
 
     /**
--- a/src/Commands/Command.php
+++ b/src/Commands/Command.php
@@ -51,6 +51,11 @@ abstract class Command implements CommandInterface
     protected $update;
 
     /**
+     * @var array|null List of bots' name which this command is available for.
+     */
+    protected $bots;
+
+    /**
      * Get Command Name.
      *
      * @return string
@@ -136,6 +141,16 @@ abstract class Command implements CommandInterface
     public function getCommandBus()
     {
         return $this->telegram->getCommandBus();
+    }
+
+    /**
+     * Checks if this command is available for this bot.
+     *
+     * @return Boolean
+     */
+    public function availableFor(Api $telegram)
+    {
+        return isset($this->bots) ? in_array($telegram->getName(), $this->bots) : true;
     }
 
     /**
--- a/src/Commands/CommandBus.php
+++ b/src/Commands/CommandBus.php
@@ -161,7 +161,9 @@ class CommandBus
      */
     public function execute($name, $arguments, $message)
     {
-        if (array_key_exists($name, $this->commands)) {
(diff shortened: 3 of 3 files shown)


irazasyed:
Hi,

Can you maybe post an example of usage of this change?

parshikov:
@irazasyed To use this just specify the array ```$bots``` in your command class like
```
class StartCommand implements Command
{
   // Command to be triggered for both bots
   private $bots = ['FirstBot', 'SecondBot'];
   ...
}

class LogCommand implements Command
{
   // Command to be triggered only by /log@FirstBot message
   private $bots = ['FirstBot'];
   ...
}

class MessageCommand implements Command
{
   // Command to be triggered only by /message@SecondBot message
   private $bots = ['SecondBot'];
   ...
}
```

It would be useful for cases with several bots.
As it has been mentioned in #33, this changes can help to handle the cases like
```
/start@something_bot
/start@ANOTHER_bot
```
I work the multi-bot support out but it goes very slow, so I decided to push this changes at least. May it help you to deal with #22.

irazasyed:
Thanks for the explanation and the examples.

I see one flaw in this, Lets assume there are 3 bots in a group. Out of which only 1 of them is yours and the privacy mode is OFF, which would let Telegram keep sending your bot all messages. Which i believe is the case with #33 issue.

So now, If i don't set the `$bots` array and the user is sending the command `/start@otherbot` which was intended to be sent to `@otherbot` and not `@yourbot` but since the commands system is designed to process any incoming commands and the `$bots` array is not set, yet the line 153 in your code's else statement is triggered (which returns true), the system will end up responding when it was not supposed to do that.

See the flaw in this?

Not only this, You also end up making API calls (To get bot name) for every incoming command messages which is not really needed and quite a waste of Bandwidth.

Instead, What i think we could probably do is this:

This would only apply to the bots which have their privacy mode turned OFF and are added to the groups. So the developer would basically add his bot name in the config or set it as an env variable.

When there's an incoming command, the command bus which handles the parsing and processing. We have this line [124](https://github.com/irazasyed/telegram-bot-sdk/blob/master/src/Commands/CommandBus.php#L124) which checks if there is a bot name present in the command text or not. If there is one, then it automatically means the bot is in a group and would give us the bot name. Now this bot name could be of any other bot as well (If privacy mode is OFF), what we can do here is, Check this bot name that we got from the command text against the bot name that was added in config by the dev and then process the command accordingly. If the command was for our bot, we respond else we do nothing. Makes sense?

I had a conversation with @jonnywilliamson as well. He too sees the flaw and agrees to my point. I hope you get what i meant and can see my point too.

That said, I don't think i could merge your solution right now at this point. Even though it's a good one and i really appreciate your efforts and contribution towards the project but with that flaw, we aren't dealing in all cases. So what you can do is, Either create a new PR with the solution i just explained above or wait for me/other contributors to push the changes. Thanks!!

jonnywilliamson:
OK,

So I've made a first attempt at fixing #33 

So this IS NOT multibot support. Only preventing your bot from answering other bots commands if they happen to match the ones you already had registered,

I'm sure there's plenty of refactoring to be done.

https://github.com/jonnywilliamson/telegram-bot-sdk/tree/registerBotName

jonnywilliamson:
Pushed a much simpler update this morning. Based on @irazasyed ever eagle and clever eye!

parshikov:
@irazasyed I understand that ```null``` in ```$bots``` variable causes handling each command in chat. I consider that as backward compatibility, not flaw. Futhermore you can set this variable in the parent class which your command is inherited from.

> You also end up making API calls (To get bot name) for every incoming command

No. In line [246](https://github.com/parshikov/telegram-bot-sdk/blob/d6a87d339b6d2bdd796594b27c331658c6a7633f/src/Api.php#L246) of ```Api.php``` bot name is requested only once if not set. But you might be right that it would be better to specify the bot name in configuration.

PR #46, opened 2015-10-10
Telegram 8th Oct Updates

jonnywilliamson:
Implemented the following changes.

https://core.telegram.org/bots/api#recent-changes

October 8, 2015

Added initial channel support for bots (no Telegram clients support this at the moment, please wait for updates):
The Chat field in the Message is now of the new type Chat.
You can now pass a channel username (in the format @channelusername) in the place of chat_id in all methods (and instead of from_chat_id in forwardMessage). For this to work, the bot must be an administrator in the channel (and that‘s exactly what Telegram clients don’t support yet — adding bots as administrators coming soon).


irazasyed:
You're fast! Thanks for the PR :+1:

PR #49, opened 2015-10-21
Improve mapRelatives in BaseObject

alexsoft:
@irazasyed mapRelatives doesn't really supports initializing inner properties
So for example we get webhook update. Message is transformed to object of Message class, but 'from' and 'chat' fields are not.
This code fixes it. I am not sure if you need that continue, but second if does all the job.

### DIFF
--- a/src/Objects/BaseObject.php
+++ b/src/Objects/BaseObject.php
@@ -65,6 +65,11 @@ abstract class BaseObject extends Collection
             foreach ($relations as $property => $class) {
                 if (!is_object($data) && isset($results[$key][$property])) {
                     $results[$key][$property] = new $class($results[$key][$property]);
+                    continue;
+                }
+                
+                if ($key === $property) {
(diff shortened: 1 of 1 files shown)


irazasyed:
Hi @alexsoft,

Thanks for opening the PR.

The way `mapRelatives` works is basically, the inner properties are initialized on-demand. I don't see any point in initializing all the inner properties and mapping them when you don't even need it.

Here's an example of how it would work with a webhook setup:

```php
$update = Telegram::getWebhookUpdates();
$message = $update->getMessage(); // Maps to Message Object
$from = $message->getFrom(); // Maps to User Object
$chat = $message->getChat(); // Maps to Chat Object
```

irazasyed:
I tested and it's not much of a difference but the improvement is still good. I noticed we could also use `get()` method from the collection API and still have the object properly mapped. Thanks!

PR #60, opened 2015-11-26
Make arguments passed to call_user_func_array() a little easier to un…

jonnywilliamson:
…derstand.

From the php docs:
image

__Purely a visual thing__. I think formatting the array in this way in this PR, makes it clear that we have passed an array to the `call_user_func_array()` method.

It also makes it easier to add MORE arguments in the future if any ever need to be sent. We can just do:
`return call_user_func_array([$this->telegram, $methodName], [$params, $secondArg]);`

### DIFF
--- a/src/Commands/Command.php
+++ b/src/Commands/Command.php
@@ -188,9 +188,9 @@ abstract class Command implements CommandInterface
             }
 
             $chat_id = $this->update->getMessage()->getChat()->getId();
-            $params[] = array_merge(compact('chat_id'), $arguments[0]);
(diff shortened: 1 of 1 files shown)


irazasyed:
We could just do it like this too:

```php
$arguments[0] = array_merge(compact('chat_id'), $arguments[0]);

return call_user_func_array([$this->telegram, $methodName], $arguments);
```

It'll support more arguments and cleaner. I didn't do this because then it'll let people just pass multiple arguments to it which could break the methods in Api.

PR #66, opened 2015-11-27
Add removeCommand() and removeCommands() methods to API.

jonnywilliamson:
Added new tests to verify the new methods.

Have also added other new tests in the new suggested format for naming. I have left the other test names alone just so we can see what both look like.


irazasyed:
So nitpick seems to be doing its job and i was bombarded with email notifications :|

jonnywilliamson:
Well I'll update my snippet for tests to put a "public" visibility for each function.

As for camel case caps? Mmmm.

irazasyed:
Have to ignore camel case caps check, Let me check their docs.

jonnywilliamson:
@irazasyed How about adding the nitpick.json config file and bypassing the test directory?

Example here:
https://gist.github.com/adamwathan/bf382f05c575db4efb6c

irazasyed:
@jonnywilliamson I just added but ignoring would completely ignore all the checks. Including like the visibility one (like it suggested) and other PSR2 violations.

PR #72, opened 2015-12-01
Mostly codesniffer fixes and some formatting

jonnywilliamson:
Removes a few CS warnings and made a couple of areas in the code clearer to understand with more spacing.

### DIFF
--- a/src/Commands/CommandBus.php
+++ b/src/Commands/CommandBus.php
@@ -112,7 +112,7 @@ class CommandBus
      *
      * @param $name
      *
-     * @return $this
+     * @return CommandBus
      */
     public function removeCommand($name)
     {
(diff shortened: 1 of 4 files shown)


src/Commands/CommandBus.php
```diff
@@ -95,18 +99,26 @@ public function addCommand($command)
...
+            )
+        );
     }
 
     /**
      * Remove a command from the list.
      *
      * @param $name
+     *
+     * @return $this
```
irazasyed: Should be `@return Api`

jonnywilliamson:
Actually should it not be `@return CommandBus` ?

That's what the addcommand/addcommands/removecommands all have....

eg:

```php
  /**
     * Add a command to the commands list.
     *
     * @param CommandInterface|string $command Either an object or full path to the command class.
     *
     *
     * @return CommandBus
     */
    public function addCommand($command)
```

I'll add the commit when you confirm

irazasyed:
Yes, That's what it should say! Sorry i didn't actually notice the file name.

PR #76, opened 2015-12-06
Added new tests and created a Mocking 'Telegram Response' class

jonnywilliamson:
Looking for feedback on this.

Have never done testing before, so I'm sure there can be many improvements or better ways to do things. Please show me how! :)

### DIFF
--- a/tests/ApiTest.php
+++ b/tests/ApiTest.php
@@ -3,19 +3,16 @@
 namespace Telegram\Bot\Tests;
 
 use Telegram\Bot\Api;
-use Prophecy\Prophet;
 use Prophecy\Argument;
-use GuzzleHttp\Client;
-use GuzzleHttp\HandlerStack;
-use GuzzleHttp\Psr7\Response;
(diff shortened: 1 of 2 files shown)


irazasyed:
Looks good to me :+1:

antoniomadonna:
Good job :) Moving to a different class seems cleaner.
Maybe, I'd just rename TGMocks to something like MocksProvider, or Mocker, what you think ?
And perhaps it should be inside the Mocks dir ?

irazasyed:
Mocker sounds good :+1:

jonnywilliamson:
@antoniomadonna Yes, I'll move the class. Good idea.

I toyed around with the name for ages, but here's what I thought.

MocksProvider was too long, especially when we will be typing it alot.
Mocker....was there any risk of this being mistook for the Mockery Library?
ResponseGenerator - Too long
Generator - too ambiguous.
APIMock - ??

If everyone thinks Mocker is good I'll change it.

PR #77, opened 2015-12-07
Add a custom ruleset file for phpcs.

jonnywilliamson:
Adding this ruleset for [phpcs](http://cs.sensiolabs.org/) (do not confuse with [php-cs-fxer](https://github.com/FriendsOfPHP/PHP-CS-Fixer)) allows us to ignore some rules which don't suit for this project.

Mainly the line length requirement of 80/120 is broken in various places because of our requirement to add the @methods to the top of some classes. If we were to break the lines up it would look terrible.

If you are using phpstorm, it's very easy to have live notification of phpcs errors. 

* Firstly download/install phpcs any way that suits. In my case I just used the `phpcs.phar` file.
* Add the path to the phar file in phpstorm settings
image
* Enable the phpcs inspections, to make it acknowledge the custom ruleset, just follow the numbers in this picture:
image

Your done!

### DIFF
--- a/ruleset.xml
+++ b/ruleset.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0"?>
+<ruleset name="Selective Standard">
+    <rule ref="PSR2">
+        <exclude name="PSR1.Methods.CamelCapsMethodName.NotCamelCaps"/>
+    </rule>
+
+    <rule ref="Generic.Files.LineLength">
+        <properties>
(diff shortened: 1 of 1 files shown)


irazasyed:
Huh! Awesome. It was really annoying me those warnings. Thanks :+1:

jonnywilliamson:
Some lines still show an error because they are REALLY long, but I thought 150 was a good compromise.

It's very easy to change it though

PR #79, opened 2015-12-08
Updated some new docs. Added tests for CommandBus

jonnywilliamson:
Some new text was added to the official docs. Reflected in the code now.

Also, I'm confused again about these tests.

I wanted to create a new test class to test the public methods on the `CommandBus.php` file. As I started to write the tests I realised that some of them have already been done from the `ApiTests.php` file.

This is because a lot of the tests in the `ApiTests.php` file, accessed the commandbus object before doing some work. ie `$this->api->addCommands()` is really a helper function/alias for `$this->getCommandBus()->addCommands()`

So where should the tests live? In the api test file or the commandbus test file? I have no idea.

So at the moment about 30% of the `commandBusTests` file has duplicate tests from the api tests.

I'm sure someone will be able to enlighten me on what the best way to organise these are.

As it stands all tests are green. :)

### DIFF
--- a/tests/CommandBusTest.php
+++ b/tests/CommandBusTest.php
@@ -21,14 +21,17 @@ class CommandBusTest extends \PHPUnit_Framework_TestCase
         $this->commandBus = new CommandBus(Mocker::createApi()->reveal());
     }
 
-    public function testAddsAndReturnsInstantiatedCommands()
+    /** @test */
+    public function it_adds_a_command_and_checks_commands_can_be_returned()
     {
         $this->commandBus->addCommand(MockCommand::class);
         $commands = $this->commandBus->getCommands();
         $this->assertInstanceOf(MockCommand::class, $commands['mycommand']);
+        $this->assertCount(1, $commands);
     }
 
-    public function testAddsMultipleCommands()
+    /** @test */
+    public function it_adds_multiple_commands()
     {
         $this->commandBus->addCommands([MockCommand::class, MockCommandTwo::class]);
         $commands = $this->commandBus->getCommands();
@@ -86,7 +89,7 @@ class CommandBusTest extends \PHPUnit_Framework_TestCase
      * @test
      * @expectedException \InvalidArgumentException
      */
-    public function it_throws_exception_if_message_is_blank_text()
+    public function it_throws_exception_if_message_is_only_blank_text()
     {
         $this->commandBus->parseCommand('');
     }
--- a/tests/CommandBusTest.php
+++ b/tests/CommandBusTest.php
@@ -1,0 +1,143 @@
+<?php
+
+namespace Telegram\Bot\Tests;
+
+use Prophecy\Argument;
+use Telegram\Bot\Objects\Update;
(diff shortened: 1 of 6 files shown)


tests/CommandBusTest.php
```diff
@@ -0,0 +1,143 @@
...
+
+    public function testAddsAndReturnsInstantiatedCommands()
+    {
+        $this->commandBus->addCommand(MockCommand::class);
+        $commands = $this->commandBus->getCommands();
+        $this->assertInstanceOf(MockCommand::class, $commands['mycommand']);
+    }
+
+    public function testAddsMultipleCommands()
+    {
```
irazasyed: Why is it named like this while the other methods are named in a different way? Can we please maintain one style?

irazasyed:
Lets keep the `commandBusTests` file and remove the duplicates from the Api Tests. I don't see any point in repeating.

The only concern however is, The code coverage, Would it still be considered as covered (if we just keep tests within commandBusTests and remove from Api Tests)?

If it won't cover, Then how about we move to a trait? So we maintain one set of methods and need not have to repeat ourselves while also passing the tests for both.

tests/CommandBusTest.php
jonnywilliamson: This was done because I've copied those tests from `APITests` and that's how they were named there as they weren't "my" tests.

If we are happy with the underscore naming system then I'll update these.

irazasyed: I see. Okay!

I'm okay with underscore naming style. Lets stick to that. Please update. Thanks!

jonnywilliamson:
It does and it doesn't cover.

The tests WILL cover the actual logic in the `commandbus` class, ie `addCommands/removeCommands` etc, however if I remove the tests from the APItests file, then the only thing we aren't covering is the alias methods...which basically just get the commandBus object and call the same commands....hence the duplication.

I think it should be ok to drop these in the `APITests`....but as I said...I'm looking for guidance in this area too.

jonnywilliamson:
Ok So duplicate tests removed. Names converted to uderscore format.


Here's what I mean about the code coverage. Technically now there are some methods not hit in the `Api.php` file that where previously. eg:
image


However, all this is doing is getting the commandBus object and calling the method on that.

Those methods ARE tested and covered in the `commandBusTest.php`  class and so I think it's fair to say that we have it covered.

irazasyed:
Okay, So what about using trait method that i mentioned? We move those tests out and import them in both test files?

jonnywilliamson:
Being honest I don't really know how to design a trait.

Secondary I don't know how that would work. In one test file you need to do `$this->api->addCommand`

in another it would be `$this->commandBus->addCommand`

So in actually the tests aren't exactly duplicate, but *extremely* similar.

irazasyed:
Traits are simply classes, Quite easy! I can help you with that.

I think we could do something like:

`$this->getApi()->addCommand()` and the `getApi` would be implemented in both the test classes which would simply return an instance of Api/CommandBus classes.

Makes sense?

irazasyed:
Will implement the other things later. Thanks for the PR :)

PR #80, opened 2015-12-09
Timeout and connection timeout setting ability added

ihoru:
As I wrote in Gitter chat here are changes to give developer ability to pass request timeout to the HttpClient.

1. How to use it?
In every method that makes request (getUpdates, sendVideo, sendMessage and all other) you can pass params: http_client_timeout and http_client_connect_timeout like here:
```
...->sendMessage(
            [
                'chat_id'                     => $chatId,
                'text'                        => $text,
                'http_client_timeout'         => 1,
                'http_client_connect_timeout' => 1,
            ]
        );
```
2. In process of working on this changes I decided also to give ability to change connect_timeout (it also was constant in GuzzleHttpClient class)
3. I added $connectTimeOut variable to the end of send method for backward compatibility.
4. I created class TestGuzzleHttpClient that extends GuzzleHttpClient just for getting passed timeouts and check them in tests. I'm not sure in correct place of the file (I though to put it into tests dir, but there is just mocks now and no src or something like that...)
5. I created tests for this ability


src/HttpClients/GuzzleHttpClient.php
```diff
@@ -103,21 +104,22 @@ public function send(
...
+     * @param string $body
+     * @param array  $options
+     * @param int    $timeOut
+     * @param int    $connectTimeOut
+     * @param bool   $isAsyncRequest
      *
      * @return array
      */
-    private function getOptions(array $headers, $body, $options = [], $timeOut, $isAsyncRequest = false)
+    private function getOptions(array $headers, $body, $options = [], $timeOut, $connectTimeOut, $isAsyncRequest = false)
```
ihoru: I know It :)
It's not my fault. But I can remove "= []" after options...

ihoru: Oh, its a robot ))) haha

src/TelegramClient.php
```diff
@@ -94,6 +99,7 @@ public function prepareRequest(TelegramRequest $request)
             $request->getMethod(),
             $request->getHeaders(),
             $request->isAsyncRequest(),
+            $request->getHtpClientParams(),
```
irazasyed: Typo! Please fix.

src/HttpClients/TestGuzzleHttpClient.php
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * @return int|null
+     */
+    public function getConnectTimeOut()
+    {
+        return $this->connectTimeOut;
+    }
+
+}
```
irazasyed: Please move tests to tests directory and this is not a unit test btw. You gotta write a proper test for this.

src/HttpClients/GuzzleHttpClient.php
irazasyed: Move it to the end with default value of 10 (after `$isAsyncRequest`).

src/TelegramRequest.php
```diff
@@ -197,6 +204,31 @@ public function getParams()
...
+
+        return $this;
+    }
+
+    /**
+     * Return special HttpClient params for this request.
+     *
+     * @return array
+     */
+    public function getHtpClientParams()
```
irazasyed: Typo!

src/HttpClients/TestGuzzleHttpClient.php
irazasyed: Nevermind, Just saw what you did. Please move these methods to the parent class itself.

ihoru: OK, I'll do everything a little bit later. Thank you for review!

jonnywilliamson:
I don't like this implementation.

We are added non-documented (as per Telegram API docs) parameters to the all the methods.

Why is this not an option that can be assigned to the Guzzle client directly instead?
Something like:
`$this->api->client->setHttpTimeout($value)`

IMO this has no reason being added to the API methods.

irazasyed:
@jonnywilliamson Actually, Yeah! That makes sense.

Just like the `setAsyncRequest()` method which can make any call async if required, Timeout too can be done like that as well.

jonnywilliamson:
Updated comment above,.

irazasyed:
I see it. But it's actually accepting time out as a param which i still don't think its a good idea. It might simply confuse people as its not documented by Telegram.

A method like `setTimeout` attached before the api method being called could be used to modify the timeout while still keeping things separate.

ihoru:
OK. Reject PR. I'll change implementation

irazasyed:
Thanks!

PR #87, opened 2015-12-23
setWebhook param is array

autowp:


irazasyed:
The whole documentation has to be rewritten as per the recent changes that were made and is pending to be tagged.

irazasyed:
I'll have to keep this pending till i release the next version to avoid confusion for people using last tagged version.

autowp:
Ooops. My mistake that is docs from older version

irazasyed:
No probs! Feel free to send PR if you're able to rewrite the docs part for the master branch. Only that thing is pending and holding back the release :)

jonnywilliamson:
https://github.com/jonnywilliamson/telegram-bot-sdk/commit/e408300f11bfd03a6aa47758b195db5a6428d983

irazasyed:
@autowp Closing this PR, Since @jonnywilliamson has sent another PR with the doc changes. The HTML files are actually automatically generated, So we need not edit those. As long as the docs in `docs` directory are updated, we can apply those in HTML too. Thanks though.

PR #88, opened 2015-12-23
Prevent Fatal error: Call to a member function has() on null

autowp:
Fatal error: Call to a member function has() on null in .../vendor/irazasyed/telegram-bot-sdk/src/Api.php on line 870

Appears when is no message received. 
To trigger: just open webhook in browser


irazasyed:
Thanks!

PR #53, opened 2015-11-09
Dependency Injection in commands constructors

antoniomadonna:
Use Laravel Container Class to resolve commands type hinted dependencies.

PR #91, opened 2015-12-28
Fix Dependency Injection in command's constructors. With tests.

jonnywilliamson:
This PR [https://github.com/irazasyed/telegram-bot-sdk/pull/53] had a number of issues that I only discovered tonight.

1) The laravel service provider was looking for a config value called: `inject_command_dependencies` yet that value/key is not found in the `telegram.php` config file.

2) The name didn't make sense. We aren't asking the user do they want to inject a dependency, we can't inject anything for them....we are asking do they want their dependency automatically RESOLVED...that we CAN do.

Therefore I added the new key and changed the name to `resolve_command_dependencies`.

3) The service provider was originally registering the commands THEN checking to see should it resolve any dependencies for any commands to be added. It was too late at that stage! 

Changed the order around in the service provider so that it knows to resolve dependencies before any commands are registered.

4) New option added in the config file.

5) Tests added to check that the dependencies are being resolved if an IOC container has been set.

### DIFF
--- a/src/Laravel/TelegramServiceProvider.php
+++ b/src/Laravel/TelegramServiceProvider.php
@@ -63,13 +63,13 @@ class TelegramServiceProvider extends ServiceProvider
                 $config->get('telegram.http_client_handler', null)
             );
 
-            // Register Commands
-            $telegram->addCommands($config->get('telegram.commands', []));
-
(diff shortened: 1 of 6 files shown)


irazasyed:
Thanks :+1:

Issue #95, opened 2016-01-03
[TODO] Implement inline API functionality


ihoru:
Docs:
https://telegram.org/blog/inline-bots
https://core.telegram.org/bots/inline
https://core.telegram.org/bots/api#inline-mode

I'm going to implement this feature nearest days and send PR asap.

irazasyed:
:+1:

ihoru:
TODO:
1. Tests
2. Docs
3. Base class InlineQueryResult with children: InlineQueryResultArticle, InlineQueryResultPhoto, InlineQueryResultGif, InlineQueryResultMpeg4Gif, InlineQueryResultVideo

ihoru:
4. Tests for ChosenInlineResult https://github.com/irazasyed/telegram-bot-sdk/pull/106

PR #97, opened 2016-01-04
Inline mode implementation

ihoru:
Here is a simple implementation of bot inline mode.
I've tested this on my own bot. It worked.
If everything is Ok, I will implement tests and docs.
Question: what do you think do we need InlineQueryResult base class with children:  InlineQueryResultArticle,     InlineQueryResultPhoto,     InlineQueryResultGif,     InlineQueryResultMpeg4Gif,     InlineQueryResultVideo?
I think it is not very necessary to create them right now.

Ref TODO: https://github.com/irazasyed/telegram-bot-sdk/issues/95

### DIFF
--- a/src/Api.php
+++ b/src/Api.php
@@ -1081,4 +1081,37 @@ class Api
 
         return $this;
     }
+
+    /**
+     * Use this method to send answers to an inline query.
+     *
+     * <code>
+     * $params = [
+     *   'inline_query_id'  => '',
+     *   'results'          => [],
+     *   'cache_time'       => 0,
+     *   'is_personal'      => false,
+     *   'next_offset'      => '',
+     * ];
+     * </code>
+     *
+     * @link https://core.telegram.org/bots/api#answerinlinequery
+     *
+     * @param array     $params
+     *
+     * @var string      $params ['inline_query_id']
+     * @var array       $params ['results']
+     * @var int|null    $params ['cache_time']
+     * @var bool|null   $params ['is_personal']
+     * @var string|null $params ['next_offset']
+     *
+     * @return bool
+     */
+    public function answerInlineQuery(array $params = [])
+    {
+        if (is_array($params['results'])) {
+            $params['results'] = json_encode($params['results']);
+        }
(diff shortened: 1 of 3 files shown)


irazasyed:
Thanks for the PR. Everything looks good.

Answer to your question: InlineQueryResult base class sounds good.

ihoru:
Ok, I will do it later. But for now this code will be enough for implementing this functionality in my bot ;)

jonnywilliamson:
Hi @ihoru . 

I'm finding it hard to see how this works. Could you explain a little more?

When an inbound UPDATE comes in, for a normal command, it looks like this:

```json
{
    "update_id": 430107013,
    "message": {
        "message_id": 203,
        "from": {
            "id": 9999999,
            "first_name": "Firstname",
            "last_name": "Surname",
            "username": "blahblah"
        },
        "chat": {
            "id": 9999999,
            "first_name": "Firstname",
            "last_name": "Surname",
            "username": "blahblah",
            "type": "private"
        },
        "date": 1456265970,
        "text": "/demo"
    }
}
```

However an inline query update looks like this:

```json
{
    "update_id": 430107012,
    "inline_query": {
        "id": "53443510725146332",
        "from": {
            "id": 9999999,
            "first_name": "Firstname",
            "last_name": "Surname",
            "username": "blahblah"
        },
        "query": "test",
        "offset": ""
    }
}
```

The SDK, tries to process the inbound UPDATE with the `commandshandler` method (in my case I set the webhook to true).

image


The actual `processCommand()` method looks like this:
image


So it only can process a command if there is a `message` branch in the Update object and when THAT has a `text` branch.

Seeing as an inline query doesn't have EITHER of those two, do you use a different method to reply to an inbound inline query object?

Do you just check to see if an inbound update object has an `inline_query` branch and then manually call the `answerInlineQuery` method?

Thanks!

ihoru:
Oh, yep. You are right. There is no implementation for inline quiet in that
class,'cause I don't use it.
I am busy right now. But, hope I can fix that later in a week. Or you can
implement it by yourself and send PR.

PR #106, opened 2016-01-15
ChosenInlineResult implementation

ihoru:
Due to new API update I've implemented ChosenInlineResult.

 https://core.telegram.org/bots/inline#collecting-feedback
https://core.telegram.org/bots/api#recent-changes

And fixed issues in documentation of InlineQuery.


irazasyed:
Thanks :+1:

Issue #32, opened 2015-09-08
[TODO] Add Emoji Support and Helpers


irazasyed:
Will use this for reference: http://emoji.muan.co/

Add support such that, Simple `:smile:` would automatically map that into appropriate char from this list: https://github.com/muan/emojilib/blob/master/emojis.json

benelang:
+1

filipekiss:
@irazasyed I'll do this.

irazasyed:
@filipekiss Sounds good. Looking forward to your PR.

Issue #103, opened 2016-01-07
[idea] An extension for Laravel schedule system to send notification through telegram bot


alberto-bottarini:
I would like to implement an extension to use Telegram Bot inside my schedule command list (app/Console/Kernel.php) to allow to send notification.

My idea is to have something like this:

$schedule->command('mycommand')->twiceDaily(0, 12)->sendOutputTo('myFile')->telegramOutputTo($chatId);

If you think this is useful, I would like to create a PR when it will be ready.

irazasyed:
Hi, The idea is good and i appreciate your thought on contribution.

However, I doubt it's something everyone would be using. Since its Laravel specific and this library is not limited to just Laravel users.

But it could very well be turned into a proper separate package as an extension for laravel users, so people can require it if they would want this feature.

Feel free to create a new repo with such feature if you want and we can link it to your repo, In fact that would very well give more ideas to others to work on such extensions and make things more better for all of us. Let me know if you need my help.

alberto-bottarini:
Hi. Extension is ready. I'm going to do all final tests and then I tell you.
You can find this here: https://github.com/alberto-bottarini/laravel-telegram-event-output

I tried to make it all the more easy as possible for developers.

alberto-bottarini:
I confirm that my extension is working very well. 
I used in production env for about 1 months and it works in a perfect way. 

I will be very grateful if you publish a link to extension on your readme :smile:

irazasyed:
Done. Added to a new dedicated wiki page here: [Telegram Bot SDK Extensions](https://github.com/irazasyed/telegram-bot-sdk/wiki/Telegram-Bot-SDK-Extensions).

Once i tag the stable version of the SDK, I'll probably setup a new page on docs site and list the extension there as well. Thanks.

PR #110, opened 2016-01-20
Add Emoji Support

filipekiss:
Regarding issue #32, added a Helper class to map keywords to Emoji. 

I used the [simplemap.json](https://github.com/muan/emojilib/blob/master/simplemap.json) as the map since it made more send to me. 
I'm not confident enough on where I should put the JSON map, so I'm open to suggestions. If there's anything you think I need to change, just let me know. :)

Also, I looked for a way to add the `simplemap.json` file as a composer dependency so we wouldn't need to version it and we could just grab from [muan/emojilib](/muan/emojilib) `master` branch. If anyone knows how to do this and think it's worth changing to this approach, let me know.


filipekiss:
So my editor messed up and my indentation is all wrong. I'll be pushing the changes soon.

irazasyed:
Thanks for the PR. Please fix the issues and commit.

As far as the emoji updates are concerned, We can try submitting a PR to that repo and ask the author to add composer support.

If not that, I found this lib: https://github.com/Ranks/emojione

It has everything we need (And more than we need in fact). Check out the [methods](https://github.com/Ranks/emojione/blob/master/lib/php/src/Client.php). Looks good to me but i hope it doesn't cause any performance issues though.

I'm thinking of just decoupling all these things into an extensions based system, So anyone who want emoji support can add another package as a dependency and enable the extension. Like there's another one here: #103

filipekiss:
Yes, using a separate package as an extension would be better. I'll take a look into [Ranks/emojione](/ranks/emojione) and test to see if I ran into performance issues.

I looked at the Travis CI building and both errors are related to composer issues, nothing to do with changes I made. Not sure how can I fix that.

The @Nitpick-CI changes were already corrected.

irazasyed:
I'll push some changes to this SDK in sometime which would integrate events system (I'll probably use Laravel's Event system maybe), Once that's integrate, you can then listen for various events and build plugins around it to manipulate the data as well as integrate more better with the SDK.

Thanks for your patience and contributions.

filipekiss:
Yeah, an event/hook system would work perfectly for this. Let me know if there's anything I can do to help. :)

irazasyed:
@filipekiss I merged this PR and added emojify to a few other things (New Keyboards) as well as auto translation for updates when retrieving text.

This will be tagged in the upcoming V3 and credited.

Thank you!

filipekiss:
No problem @irazasyed :)

Glad to help!

PR #129, opened 2016-04-01
Synced docblock to match change on official bot api.

jonnywilliamson:
Very small text change was added to official docs on 30th March 2016.


irazasyed:
Thank you!

PR #133, opened 2016-04-13
The groundwork for the future Telegram Bot API update

ihoru:
Try to make post request in method was not found.


irazasyed:
Thanks!

PR #142, opened 2016-04-16
Fix issue with array to string conversion when trying to use

jonnywilliamson:
forceReply and hide keyboard.

### DIFF
--- a/src/Keyboard/Keyboard.php
+++ b/src/Keyboard/Keyboard.php
@@ -168,7 +168,7 @@ class Keyboard extends Base
      */
     public function hide(array $params = [])
     {
-        return $this->items = array_merge(['hide_keyboard' => true, 'selective' => false], $params);
+        return $this->merge(array_merge(['hide_keyboard' => true, 'selective' => false], $params));
     }
 
     /**
@@ -192,6 +192,6 @@ class Keyboard extends Base
      */
     public function forceReply(array $params = [])
     {
-        return $this->items = array_merge(['force_reply' => true, 'selective' => false], $params);
+        return $this->merge(array_merge(['force_reply' => true, 'selective' => false], $params));
     }
 }


ahmadbadpey:
when I changes the lines of file  to : 

`return $this->merge(array_merge(['hide_keyboard' => true, 'selective' => false], $params));`

I got this error :  
`$this is not accessible in static context`

irazasyed:
@ahmadbadpey Why would you do that? Please use `dev-develop` branch. Also, create a new issue if you have any issues. Thanks!

ahmadbadpey:
@irazasyed , I tried to upgrade to `dev-develop` mentioned by you, but after that first I got this error : 
```
Cannot use Telegram\Bot\Objects\Message as Message because the name is already in use {"exception":"[object]  
 (Symfony\\Component\\Debug\\Exception\\FatalErrorException(code: 64): Cannot use Telegram\\Bot\\Objects\\Message as Message because the name is already in use at D:\\wamp\\www\\botshop\\vendor\\irazasyed\\telegram-bot-sdk\\src\\Methods\\Payments.php:5)
```

And then when I removed `use Telegram\Bot\Objects\Message; ` from `payments.php` now I got this error : 

`Declaration of App\Commands\StartCommand::handle($arguments) must be compatible with Telegram\Bot\Commands\Command::handle()`

ahmadbadpey:
No I found that if I use below approach to make a fore_reply button , problem solves : 

```
$forceReply = Keyboard::forceReply(['force_reply' => false]);
$this->replyWithMessage([
		'text'         => 'Please enter your name : ',
		'reply_markup' => $forceReply
]);
```

Before I used this : 

```
$reply_markup = [
		'force_reply' => true
];
$this->replyWithMessage([
		'text'         => 'Please enter your name : ',
		'reply_markup' => $reply_markup 
]);
```

PR #143, opened 2016-04-16
Fix Emoji parsing.

jonnywilliamson:
When writing text, the emoji keyword must be **enclosed** with a colon 
eg
 `:smile:`

When converting from emoji back to text, the text word returned will also be enclosed by colons
 eg
Before:
 `That is great news! smile`
After:
 `That is great news! :smile:`


irazasyed:
Thanks @jonnywilliamson :)

PR #145, opened 2016-04-17
answerCallbackQuery method support

skythet:


irazasyed:
Thanks for the PR but this method already exists.

PR #151, opened 2016-04-21
Allow updates to be processed for commands manually.

jonnywilliamson:
As discussed the in the slack channel. Provides a way to process a command manually, ie using a queue system etc


irazasyed:
Thanks!

PR #154, opened 2016-04-21
Fix for object being passed instead of array

jonnywilliamson:
https://github.com/irazasyed/telegram-bot-sdk/commit/ae214d081928afd4dc9916493768158c28075790#commitcomment-17194269

### DIFF
--- a/src/Laravel/TelegramServiceProvider.php
+++ b/src/Laravel/TelegramServiceProvider.php
@@ -74,7 +74,7 @@ class TelegramServiceProvider extends ServiceProvider
         $app->singleton('telegram', function ($app) {
             $config = $app['config']['telegram'];
 
-            return (new BotsManager($config))->setContainer($app);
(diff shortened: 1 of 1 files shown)


src/Laravel/TelegramServiceProvider.php
```diff
@@ -74,7 +74,7 @@ protected function registerManager(Application $app)
         $app->singleton('telegram', function ($app) {
             $config = $app['config']['telegram'];
 
-            return (new BotsManager($config))->setContainer($app);
+            return (new BotsManager((array)$config))->setContainer($app);
```
jonnywilliamson: Or should the cast to array happen on line 75?

irazasyed: Yes

jonnywilliamson:
Cast to array placed when the `$config` if being created.

jonnywilliamson:
Thank you!

PR #157, opened 2016-04-24
~1k speed up word to emoji replace using regexp

halaei:


src/Helpers/Emojify.php
```diff
@@ -158,9 +158,13 @@ protected function replace($line, $replace, $toWord = false, $delimiter = ':')
      */
     protected function wordToEmojiReplace($line, $replace, $delimiter)
     {
-        foreach ($replace as $key => $value) {
-            $line = str_replace($delimiter.$key.$delimiter, $value, $line);
-        }
+        $pattern = '/' . preg_quote($delimiter, '/') . '([^:]*)' . preg_quote($delimiter, '/') . '/';
```
ihoru: You've forgotten to put here `[^:]` calling `preg_quote($delimiter, '/')` as well, I think.

halaei: Yes @ihoru . Thanks

irazasyed:
Thanks for the PR!

Issue #158, opened 2016-04-24
emojify cannot be completely dissabled via config


halaei:
`Emojify::text()` is called in `Telegram\Bot\Keyboard\Button::setText()` for example, with no option of dissabling it.

PR #159, opened 2016-04-24
optionally disabling emojify

halaei:
#158


irazasyed:
Thanks for the PR but I'll probably look for ways to actually disable it from the main class instead of doing it from outside the class like now.

irazasyed:
I'll keep this open until i push my solution or feel free to create a new PR with your solution if you have any.

halaei:
Thanks. Please note that `Emojify::translate()` is still slow

PR #160, opened 2016-04-24
Small tweak to word to emoji parsing regex.

jonnywilliamson:
Small tweak with wordToEmoji parsing regex.


ihoru:
Thanks ;)

jonnywilliamson:
I agree with earlier comment. Regex should use a + and not a *

With asterisk, we match double colons:

image


With Plus we do not

image

irazasyed:
Thank you!

PR #162, opened 2016-04-25
Revet regexp

halaei:
I think my regexp is not working and it is impossible to solve this problem efficiently with php implementation of regexp functions.
Here is what that is wrong:

input:
':foo:one:foo'
expected output:
':foo1foo:'
current output:
'fooonefoo'


jonnywilliamson:
Wait what?

I think github's parser has messed up the problem case.

Could you repeat the issue again 'using code tags?`

halaei:
@jonnywilliamson my comment is now updated

Issue #168, opened 2016-04-27
Why method getName is not a part of CommandInterface?


BoShurik:
So, we should add this method to interface or make a check for `\Telegram\Bot\Commands\Command` instance

https://github.com/irazasyed/telegram-bot-sdk/blob/master/src/Commands/CommandBus.php#L102

irazasyed:
Feel free to send a PR. I didn't add. But i will later, thanks for bringing that up.

PR #170, opened 2016-04-27
Update CommandInterface

BoShurik:
Fix https://github.com/irazasyed/telegram-bot-sdk/issues/168

Also remove `hande` method as it used only in `\Telegram\Bot\Commands\Command` and can not be executed without `make` method (because `telegram` and `update` properties are initialized there)

And add type hinting to make method


irazasyed:
Thank you!

Issue #90, opened 2015-12-27
[WIP] Add Conversational System


irazasyed:
I have plans to create a system with which, we can build a Q/A kinda interactive bots easily.

Since the commands system is not designed to handle this nor it was intended to do that, A separate functionality makes sense.

The only part that's quite complex is to track the answers/responses and to map it to the right question and then move to next in a queue.

Any inputs would be good.

filipekiss:
Telegram itself is not prepared to handle conversations, actually. There's no continuity when a user talks to a bot. This has to be done on application side. I use this on my Package Tracking Bot (when user is adding a package, he asks for the package description and etc) and I log what that user last command was and what the next step should be. I can try to decouple it from my app the best I can, but you'd still need a way to record the user current step so you can decided the next one (I currently use a SQLITE database for that).

irazasyed:
That's right. We were working on one such project with same requirements and we ended up using database method to keep track.

But will leave this open to try someday with maybe some other option like Cache or such.

filipekiss:
Yeah, it's kinda hard because even though it's not that hard to do, it's still mostly living inside the application and not on the framework itself...

irazasyed:
# Update

Thanks to @nMado who offered to work on this feature. We've been discussing on how to implement this for the last few days on [PHP Chat](https://phpchat.co) inside a private channel.

He'll be posting a TODO list of things that we'll be working on to implement into this project soon.

If anyone else want to join the discussion and contribute, feel free to join [PHP Chat](https://phpchat.co) Slack Community. Once you do, Please Direct Message (DM) me or post in #hangout channel that you want to join `telegram-bot-sdk` private channel and I'll invite you there. Thanks!

P.S Excited to see this feature implemented. Thanks again @nMado for coming forward to contribute to this project :)

nMado:
Thank you @irazasyed .

TODO:

  * [ ] Build conversational base.
  	* [ ] `ConversationBase` class.
  		(Detects conversational commands, starts first question or continues from where the user left, manages cache ...)

    * [ ] `ConversationQuestion` class.
    	(Represents a question, contains a unique name for the question, holds the answer callback ...)

    * [ ] Use cache to store answers.

  * [ ] Implement fluent API in the `Command` class. (i.e. `$this->startConversation()`)
  * [ ] Make sure it supports laravel and non-laravel users.
  * [ ] Write tests.

[...]

PR #171, opened 2016-04-29
Add conversational system (#90)

tjomamokrenko:
Bot example:
```php
<?php

require_once 'vendor/autoload.php';

$telegram = new \Telegram\Bot\Api('secret');

$telegram->getConversationBus()->addConversations([
    \Telegram\Bot\Conversations\HelloConversation::class,
    \Telegram\Bot\Conversations\LeaveMeAloneConversation::class,
]);

$telegram->conversationsHandler(false);
```

Conversation example:
```php
<?php


namespace Telegram\Bot\Conversations;


/**
 * Class HelloConversation
 * @package Telegram\Bot\Conversations
 * @author Artiom Mocrenco <tjomamokrenko@gmail.com>
 */
class HelloConversation extends Conversation
{
    /**
     * @inheritdoc
     */
    public function handle()
    {
        $text = \strtolower($this->getUpdate()->getMessage()->getText());

        if ((strpos($text, 'hi') !== false) || (strpos($text, 'hello') !== false)) {
            $this->replyWithMessage(['text' => "Hi, {$this->getUpdate()->getMessage()->getFrom()->getFirstName()}!"]);
            
            $this
                ->getUpdate()
                ->getMessage()
                ->getFrom()
                ->setCurrentConversation(
                    LeaveMeAloneConversation::class
                );
        }
    }
    
}
```

screenshot_2016-04-29-03-24-12

Not ready for production until:
- laravel support is added
- a small refactoring is done
- tests are written

Your thoughts? :smiley:


tjomamokrenko:
https://github.com/irazasyed/telegram-bot-sdk/issues/90

bot.php
```diff
@@ -1,12 +0,0 @@
-<?php
-
-require_once 'vendor/autoload.php';
-
-$telegram = new \Telegram\Bot\Api('141984924:AAGWpUpjFjBfTo323D_-nDDOxaRxVcdUiB0');
```
tjomamokrenko: oops :smile:

src/Objects/User.php
```diff
@@ -2,6 +2,12 @@
 
 namespace Telegram\Bot\Objects;
 
+use bar\baz\source_with_namespace;
```
irazasyed: Is this a valid namespace? Doesn't look like.

src/Objects/User.php
```diff
@@ -14,6 +20,57 @@
...
+    {
+        if (is_null($this->currentConversation)) {
+            $cacheId = $this->getId() . '_current_conversation';
+            if (($currentConversation = $this->cache->fetch($cacheId)) !== false) {
+                $this->currentConversation = $currentConversation;
+
+                return $currentConversation;
+            };
+
+            $this->setCurrentConversation(HelloConversation::class);
```
irazasyed: Why are we setting a default conversation?

src/Objects/User.php
```diff
@@ -14,6 +20,57 @@
...
+    private $cache;
+
+    /**
+     * @inheritdoc
+     */
+    public function __construct($data)
+    {
+        parent::__construct($data);
+
+        $this->cache = new FilesystemCache(__DIR__.'/../Storage/cache');
```
irazasyed: I think we could've used `illuminate/cache` and support multiple cache drivers?

For example, I prefer redis or memcached over a file cache.

src/Conversations/HelloConversation.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Telegram\Bot\Conversations;
+
+/**
+ * Class HelloConversation
+ * @package Telegram\Bot\Conversations
+ * @author Artiom Mocrenco <tjomamokrenko@gmail.com>
+ */
```
irazasyed: Please remove package and author. You'll be credited in release notes. We need to be consistent.

src/Conversations/LeaveMeAloneConversation.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Telegram\Bot\Conversations;
+
+/**
+ * Class HelloConversation
+ * @package Telegram\Bot\Conversations
+ * @author Artiom Mocrenco <tjomamokrenko@gmail.com>
+ */
```
irazasyed: Please remove package and author. You'll be credited in release notes. We need to be consistent.

src/Answers/Answer.php
```diff
@@ -0,0 +1,76 @@
...
+ * @method mixed replyWithAudio($use_sendAudio_parameters)           Reply Chat with an Audio message. You can use all the sendAudio() parameters except chat_id.
+ * @method mixed replyWithVideo($use_sendVideo_parameters)           Reply Chat with a Video. You can use all the sendVideo() parameters except chat_id.
+ * @method mixed replyWithVoice($use_sendVoice_parameters)           Reply Chat with a Voice message. You can use all the sendVoice() parameters except chat_id.
+ * @method mixed replyWithDocument($use_sendDocument_parameters)     Reply Chat with a Document. You can use all the sendDocument() parameters except chat_id.
+ * @method mixed replyWithSticker($use_sendSticker_parameters)       Reply Chat with a Sticker. You can use all the sendSticker() parameters except chat_id.
+ * @method mixed replyWithLocation($use_sendLocation_parameters)     Reply Chat with a Location. You can use all the sendLocation() parameters except chat_id.
+ * @method mixed replyWithChatAction($use_sendChatAction_parameters) Reply Chat with a Chat Action. You can use all the sendChatAction() parameters except chat_id.
+ */
+abstract class Answer
```
irazasyed: Turn this into a trait? So that it can be used anywhere.

Possible names: `CanReply` / `Replyable` / `Answerable` / `ReplyWith`

irazasyed:
Thanks for the PR. Please review Nitpick & my comments and make the changes.

src/Objects/User.php
tjomamokrenko: @irazasyed It's like a starting point. But it definetely does not to be hardcoded.
I see two ways of solving this:
- Set default (or starting) conversation elsewhere in application (e.g. in standalone or laravel bot)
- Let users overwrite ConversationBus::handler() method

src/Objects/User.php
tjomamokrenko: We can use illuminate/cache if you think it is better. I've thought about the default cache driver to be SQLite3 or other than file, but they are not supported on every system. So I started with filesystem cache, which is supported. Of course multiple cache drivers support is on the way since if we have more than 10000 users having conversation than more than 10000 cache files are created, which is definetely a problem. I chose `doctrine/cache` because it's simpler, well-tested and more stable.

So,
1. Are you sure you want me to rewrite it using `illuminate/cache`?
2. What cache driver support to implement next? SQLite3, Redis, Memcached or other?

src/Conversations/HelloConversation.php
tjomamokrenko: OK, sure, I will remove them in next commits

src/Conversations/LeaveMeAloneConversation.php
tjomamokrenko: OK, sure, I will remove them in next commits

src/Objects/User.php
irazasyed: I think the first one makes sense. But what happens if i don't want to use the whole conversation system?

Also, If I ask a question to the user and I'm expecting an answer, how would you handle that in this conversation ? I'd like to see an example.

Example questions:
- What is your age? (Bot is expecting an integer value between 18-25).
- User responds: twenty years old
- Bot replies: Invalid Answer, Please tell me in numeric value.
- User responds: 17
- Bot replies: Sorry, Only people who're older than 18 can use this bot.
- User responds: 19
- Bot replies: Thank you.
- Bot sends next question: Where are you from?
...

Conversation continues.

src/Objects/User.php
tjomamokrenko: Yes, just PHPStorm included it by mistake. Will be removed.

src/Objects/User.php
irazasyed: I'm not quite sure if it's any better because I haven't really used `doctrine/cache` directly. I've only used `illuminate/cache` in Laravel projects.

The problem with `illuminate` components is basically when the project is used within some incompatible version of Laravel, then we're forced to update the changes to our project as well as maintain multiple versions just to support Laravel users, which is kinda problem.

If `doctrine/cache` can support multiple drivers and it's easier, then I don't have any issues going with it.
The default could be file cache while giving users a choice to switch to Redis / Memcached / DB (SQLite or MySQL if that's possible).

Can we detect the platform and maybe use Laravel's built-in cache and settings when it's being used with Laravel and fallback to `doctrine/cache` for standalone users? That would make things a lot easier.

src/Objects/User.php
irazasyed: Ok cool.

src/Conversations/LeaveMeAloneConversation.php
irazasyed: Thanks!

src/Conversations/HelloConversation.php
irazasyed: Thanks!

src/Objects/User.php
tjomamokrenko: 1. If you don't want to use conversation system, than you can disable it (has to be implemented). Or, we can make it disabled by default for now. Laravel users will be able to enable/disable it in telegram.php, and standalone will just need to call a method.
2. I will respond you with a solution today.

irazasyed: Cool. Looking forward to it.

src/Conversations/ConversationInterface.php
```diff
@@ -0,0 +1,26 @@
...
+     * @return mixed
+     */
+    public function make(Api $telegram, Update $update);
+
+    /**
+     * Handle Conversation.
+     *
+     * @return mixed
+     */
+    public function handle();
```
BoShurik: I think this method is redundant. It only used in AbstractCommand and should not be public

src/Objects/User.php
tjomamokrenko: 1. The question about cache provider remains open, we will decide today or when you have time.
2. Using laravel cache with it's settings, in my point of view, brings magic and unexpected results (e.g. reserved cache keys for telegram-bot-sdk) for developers. 
3. What would you say about this order of implementation of cache drivers support: Filesystem -> SQLite3 -> Redis -> Memcached?

src/Conversations/ConversationInterface.php
tjomamokrenko: @BoShurik You are right.

src/Objects/User.php
irazasyed: 1. Ok.
2. Not sure I get that.
3. The order is good and makes sense.

BoShurik: IMHO, we should not care about implementation. Maybe just rely on the interface:
``` php
use Doctrine\Common\Cache\Cache

public function __construct($data, Cache $cache = null)
```

tjomamokrenko:
- [ ] Remove invalid namespace
- [ ] Set default conversation elsewhere in application
- [ ] Add ability to disable conversational system
- [ ] Choose between `illuminate/cache` and `doctrine/cache` (https://github.com/irazasyed/telegram-bot-sdk/issues/172)
- [ ] Rely on interface when caching conversation data
- [ ] Remove package and author entries from docblocks
- [ ] Turn abstract class Answer into a trait (?)
- [ ] Remove redundant handle() method from interface
- [ ] Apply style fixes

ihoru:
I don't like all this stuff being in this project. This implementation should be done on another layer, not in Api class...
The best choice would be to create separate library with pure Api requests and some related helpers. And this SDK library will require it.

tjomamokrenko:
@ihoru Well, this is possible. What's your opinion? @irazasyed

tjomamokrenko:
@ihoru You are right, it will be more convenient to create another composer package. 

However, I will continue to contribute to this package, but not regarding conversations.

Look here. https://github.com/tjomamokrenko/telegram-bot-conversation/tree/develop

irazasyed:
@tjomamokrenko I'll do one thing. I'll create a new org on github, Start decoupling things into packages and we can add the conversation as one separate package there as well. What say?

cc: @ihoru @jonnywilliamson

irazasyed:
Here you go: https://github.com/TelegramBotKit/conversation

Right now I've added @tjomamokrenko for that repo. Anyone else interested to handle something, please let me know and I'll add you to the appropriate repo.

I'll be adding SDK and other things soon.

PR #175, opened 2016-04-29
Use events to notify about updates

tjomamokrenko:
Because `sdk` will be decoupled on different packages and they will need to receive updates from it.

Usage example (advanced):
```php
<?php

use League\Event\Emitter;
use Telegram\Bot\Api;
use Telegram\Bot\Events\UpdateReceivedEvent;

$telegram = new Api('secret');

$telegram->setEventEmitter(new Emitter());

$telegram
    ->getEventEmitter()
    ->addListener('Telegram\Bot\Events\UpdateReceivedEvent', function ($event) {
    /** @var UpdateReceivedEvent $event */

    // do something here
    assert($event->getUpdate() instanceof \Telegram\Bot\Objects\Update);

}, Emitter::P_NORMAL);

/** @var \League\Event\ListenerInterface $handleCommandsListener */
$telegram
    ->getEventEmitter()
    ->addListener('Telegram\Bot\Events\UpdateReceivedEvent', $handleCommandsListener);

/** @var \League\Event\ListenerInterface $handleConversationsListener */
$telegram
    ->getEventEmitter()
    ->addListener('Telegram\Bot\Events\UpdateReceivedEvent', $handleConversationsListener);

/** @var \League\Event\ListenerInterface $userDefinedListener */
$telegram
    ->getEventEmitter()
    ->addListener('Telegram\Bot\Events\UpdateReceivedEvent', $userDefinedListener);

// events will be emitted
$telegram->getUpdates();

// events will not be emitted
$telegram->getUpdates([], false);

// event will be emitted
$telegram->getWebhookUpdates();

// event will not be emitted
$telegram->getWebhookUpdates(false);
```

If you don't want to listen to the events:
```php
<?php

use Telegram\Bot\Api;

$telegram = new Api('secret');

// events will not be emitted, because emitter is not present
$telegram->getUpdates();

// events will not be emitted
$telegram->getUpdates([], false);

// event will not be emitted, because emitter is not present
$telegram->getWebhookUpdates();

// event will not be emitted
$telegram->getWebhookUpdates(false);
```

### DIFF
--- a/src/Api.php
+++ b/src/Api.php
@@ -5,10 +5,9 @@ namespace Telegram\Bot;
 use Illuminate\Contracts\Container\Container;
 use Telegram\Bot\Commands\CommandBus;
 use Telegram\Bot\Events\EmitsEvents;
-use Telegram\Bot\Events\UpdateReceivedEvent;
+use Telegram\Bot\Events\UpdateWasReceivedEvent;
 use Telegram\Bot\Exceptions\TelegramSDKException;
 use Telegram\Bot\FileUpload\InputFile;
-use Telegram\Bot\HttpClients\GuzzleHttpClient;
 use Telegram\Bot\HttpClients\HttpClientInterface;
 use Telegram\Bot\Objects\File;
 use Telegram\Bot\Objects\Message;
@@ -962,14 +961,14 @@ class Api
      *
      * @return Update
      */
-    public function getWebhookUpdates($emitUpdateReceivedEvent = true)
+    public function getWebhookUpdates($emitUpdateWasReceivedEvent = true)
     {
         $body = json_decode(file_get_contents('php://input'), true);
 
         $update = new Update($body);
 
-        if ($emitUpdateReceivedEvent) {
-            $this->emitEvent(new UpdateReceivedEvent($update));
+        if ($emitUpdateWasReceivedEvent) {
+            $this->emitEvent(new UpdateWasReceivedEvent($update));
         }
 
         return $update;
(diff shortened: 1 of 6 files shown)


src/Api.php
```diff
@@ -4,9 +4,10 @@
 
 use Illuminate\Contracts\Container\Container;
 use Telegram\Bot\Commands\CommandBus;
+use Telegram\Bot\Events\EmitsEvents;
+use Telegram\Bot\Events\UpdateWasReceivedEvent;
```
irazasyed: It could just say `UpdateWasReceived`, No need to affix with `Event`. Since the namespace itself is different and it already gives an idea of it being an event.

src/Api.php
```diff
@@ -4,9 +4,10 @@
...
 use Telegram\Bot\Exceptions\TelegramSDKException;
 use Telegram\Bot\FileUpload\InputFile;
-use Telegram\Bot\HttpClients\GuzzleHttpClient;
```
irazasyed: Why was this removed? Not being used?

tjomamokrenko: @irazasyed Yes.

tjomamokrenko:
@irazasyed Here's an example of listener in talk package https://github.com/telegram-bot-kit/talk/blob/develop/src/Listeners/HandleConversationsListener.php

src/Api.php
```diff
@@ -958,11 +961,17 @@ public function setWebhook(array $params)
      *
      * @return Update
      */
-    public function getWebhookUpdates()
+    public function getWebhookUpdates($emitUpdateWasReceivedEvent = true)
```
irazasyed: I think we could have a global config option to enable / disable events. Because the events could be applied for things like message sending, photo sending, etc. And we can't really have params for each of them to turn on/off. It would make sense to just enable / disable from the main.

tjomamokrenko: @irazasyed We need to have it, BUT the purpose of this param is the following:
Imagine if getWebhookUpdates is called twice... then events are emitted twice. What if developer does want it to happen just once? That's the purpose.

src/Api.php
tjomamokrenko: @irazasyed Fix coming.

PR #177, opened 2016-05-01
Move type detection from Api class to Message and Update.

Sloothword:
In line with our chat discussion:
- Detection of message type should be concern of Message object --> Moved there
- Detection of Update object added
- Made Api methods more robust against error or updates without message
- Deprecated Api methods

Am i doing this right? (first time github)

Questions:
- Why is `$message->isType` not just an alias for `$message->has`. What does the extra `return $this->detectType() === $type;` add?
- `Api->detectMessageType` was (and is) untested. I honestly do not really know how to test, cause in the end you only test if $update->keys() is implemented correctly. Maybe we need some recorded(?) telegram updates to use as test input.

### DIFF
--- a/src/Api.php
+++ b/src/Api.php
@@ -1180,64 +1180,60 @@ class Api
     /**
      * Determine if a given type is the message.
      *
+     * @deprecated Call method isType directly on Message object
+     *             To be removed in next major version.
+     * 
      * @param string         $type
      * @param Update|Message $object
      *
+     * @throws \InvalidArgumentException
+     * 
      * @return bool
      */
     public function isMessageType($type, $object)
     {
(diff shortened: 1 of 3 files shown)


irazasyed:
Thanks for your contribution :)

You're doing it right.

Answer to your questions:
1. It helps to guess what type of message it is. There was a requirement sometime ago by a user of this project and it made sense back then to create like that. Apart from that, I don't remember why i had gone with that method TBH. But I'm sure I had added for a valid reason.
2. Yes, It's untested. In fact a lot of tests are missing, Since we're decoupling the project, I'm not writing any tests right now. Will be looking at it at the end.

PR #182, opened 2016-05-21
[WIP] Move non-API concerns out of Api

sloothword:
I believe we should seperate the concerns of the Api class, which have nothing to do with a barebones telegram Api.
Like handling commands, emitting events, building conversations, etc.
For these things i created a separate Bot class.

Things still needed to be done before merging:

- [ ]  Understand, implement and test the container/DI stuff
- [X] Document the Bot class

- [ ] Write a usage example of `Bot`, `checkForUpdates` and `Events`.
Is there an official Event tutorial? (I know the one in the pull request, but thats not official documentation) How to update the docs? What is the status of the upgrade guide?

- [X] Fix current tests

- [x] Try to write some tests. (As i mainly moved some untested stuff from one place to another, there is not much to copy/modify) I am open for test suggestions ;-)

Things to discuss (and tackle maybe afterwards)

- [ ] Is all the magic `__call` stuff helpful or confusing. Yes, it may be convenient to call everything everywhere, but it also introduces bugs (the `getXXX` in Api) and adds complexity.

- [ ] The variable holding the telegram api object is not consistently named. Sometimes `$telegram`, sometimes `$api`. It gets a bit better with the distinction of `$bot`, but imho it should always be `$api`, never `$telegram`.

- [ ] The config system is a bit too intricate. I get what you are trying to do with reusable config blocks like the shared commands and such, but with the addition of e.g. the event and conversation system we might find a seamless config mechanism for all of that.

### DIFF
--- a/src/Answers/AnswerBus.php
+++ b/src/Answers/AnswerBus.php
@@ -2,6 +2,7 @@
 
 namespace Telegram\Bot\Answers;
 
+use Telegram\Bot\Bots\Bot;
 use Telegram\Bot\Api;
 
 /**
@@ -10,9 +11,9 @@ use Telegram\Bot\Api;
 abstract class AnswerBus
 {
     /**
-     * @var Api
+     * @var Bot
      */
-    protected $telegram;
+    protected $bot;
 
     /**
      * Handle calls to missing methods.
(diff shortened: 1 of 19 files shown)


irazasyed:
Thanks for the PR but please follow the contributing guidelines. There are a few issues with this. Things like coding style, etc.

sloothword:
Yeah of course. I meant it as a WIP, an idea how it could look like (and maybe start a discussion about the unclear topics)
If the core intent is ok with you i happily fix all the issues (some are even listed in the PR itself).

Just wanted to check before i invest more time in something which potentially gets scrapped.

irazasyed:
Yeah, It looks good. More decoupling will be done before V3 and moved to the new org.

Issue #204, opened 2016-07-07
not work replyWith in callback data with inlinekeyboard 


Ebi1020:
hi, i'm get error in commands :(
[2016-07-08 03:57:48] local.ERROR: exception 'Symfony\Component\Debug\Exception\FatalErrorException' with message 'Call to a member function getChat() on null' in H:\src\vendor\irazasyed\telegram-bot-sdk\src\Answers\Answerable.php:52

**but** working with edit line 52 to:

if ($this->update->has('message'))
        $chat_id = $this->update->getMessage()->getChat()->getId();
else
        $chat_id = $this->update->getCallbackQuery()->getMessage()['chat']->getId();

biakaveron:
The same issue.

I think it will be better to implement getChat() method for Update object.

jonnywilliamson:
This is because the sdk has an assumption that all Updates passed to this class will be of the `message` type when in fact they can also be `edited_message` type.

We are working on a solution to this.

jonnywilliamson:
Hopefully this should be done shortly.

korvin3:
I have the same problem. My solution:
`$telegram->getCommandBus()->execute('*callbackName*', [], $update->getCallbackQuery());`
Maybe be useful for someone.

PR #205, opened 2016-07-18
PR for broken replyWith + callback data

biakaveron:
See #204


jonnywilliamson:
Do we not also have to deal with the `edited_message` type here as well?

irazasyed:
@jonnywilliamson Will implement that along with other changes as per the develop branch.

irazasyed:
Thanks @biakaveron

PR #217, opened 2016-08-06
add answerCallbackQuery request method

SamsamBabadi:
added https://core.telegram.org/bots/api#answercallbackquery request for answer CallbackQuery

### DIFF
--- a/src/Api.php
+++ b/src/Api.php
@@ -1472,4 +1472,32 @@ class Api
 
         return $this;
     }
+    
+    /**
+     * Answer Callback Query.
+     *
+     * <code>
+     * $params = [
+     *   'callback_query_id'        => '',
+     *   'text'                     => '',
+     *   'show_alert'               => '',
+     * ];
+     * </code>
+     *
+     * @link https://core.telegram.org/bots/api#answercallbackquery
+     *
+     * @param array    $params
+     *
+     * @var string     $params ['callback_query_id']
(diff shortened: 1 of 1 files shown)


jonnywilliamson:
This already exists in the [Api.php](https://github.com/irazasyed/telegram-bot-sdk/blob/master/src/Api.php#L757-L783) file. However it's slightly different:

```php

    public function answerCallbackQuery(array $params)
    {
        return $this->post('answerCallbackQuery', $params);
    }

```

Should this return a TelegramResponse or a new Message object?

I haven't done enough testing on this to decide.

irazasyed:
@SamsamBabadi Thanks for opening a PR.

Like @jonnywilliamson already mentioned, This method is already implemented and looking at the docs, It seems like on success it simply returns a boolean value. Hence, just returning TelegramResponse makes sense than a new Message object -- Perhaps we could return a decoded body too.

SamsamBabadi:
@jonnywilliamson & @irazasyed Thanks for your answers.
I think TelegramResponse is good.
I open this PR because I didn't see this method in last released version. :D
Forever I try to use dev-master version for used this method but none of the methods does not work.

irazasyed:
@SamsamBabadi You'll have to pull `3.*@dev` to get the dev-master version. I'm closing this PR then. Thanks anyway 👍

PR #218, opened 2016-08-10
Implemented recent Telegram API updates.

jonnywilliamson:
Implemented recent Telegram API updates.

May 22, 2016

Bot API 2.1. Added more tools for group administrator bots. Your bot can now get a list of administrators and members count in a group, check a user's current status (administrator, creator, left the group, kicked from the group), and leave a group.
Added new methods: getChat, leaveChat, getChatAdministrators, getChatMember, getChatMembersCount.
Added support for edited messages and new mentions from Telegram v.3.9. New fields: edited_message in Update, edit_date in Message, user in MessageEntity. New value text_mention for the type field in MessageEntity.

May 6, 2016

Added the field emoji to the Sticker object. Your bot can now know the emoji a sticker corresponds to.
Added the field forward_from_chat to the Message object for messages forwarded from channels.

### DIFF
--- a/src/Api.php
+++ b/src/Api.php
@@ -9,6 +9,8 @@ use Telegram\Bot\Events\UpdateWasReceived;
 use Telegram\Bot\Exceptions\TelegramSDKException;
 use Telegram\Bot\FileUpload\InputFile;
 use Telegram\Bot\HttpClients\HttpClientInterface;
+use Telegram\Bot\Objects\Chat;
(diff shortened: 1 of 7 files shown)


irazasyed:
Thank you 👍

jonnywilliamson:
Ugh. I see some doc block mistakes.

Always after I've pushed! I'll fix shortly.

PR #220, opened 2016-08-11
Make all api method return types consistent with the official API docs.

jonnywilliamson:
Make all api method return types consistent with the official API docs.


jonnywilliamson:
Some methods moved in the class to reflect the order of the official api docs.

PR #224, opened 2016-08-20
Code review and docblock updates

jonnywilliamson:
Update method docblocks to show new InputFile method of uploading/sending files/resources to telegram

Extract code into its own method for better readability.


irazasyed:
Thanks!

Issue #261, opened 2016-10-05
Get text of the message (for v3)


kettari:
Hi,

Am I the only one who can't understand how to get text of the message in v3?

Here #207 guy asked about v2 — it's OK with Message::getText(). But in v3 I didn't found any property for this in the [Message class](https://github.com/irazasyed/telegram-bot-sdk/blob/develop/src/Objects/Message.php).

Thanks in advance for your help.

PR #262, opened 2016-10-05
Added $text property to the Message class + fixed $game desc

kettari:
See #261


irazasyed:
Thanks for creating this PR 👍

PR #268, opened 2016-10-17
Drop PHP 5.5 support

vdavari:
Hi..

As you can see PHP 5.5 is already outdated and it's time to drop supporting this version.

http://php.net/supported-versions.php

Regards,


irazasyed:
Thanks 👍

PR #280, opened 2016-11-01
Fix Php7 Invalid Class Name Issue

jonnywilliamson:
Prevent conflict with reserved key word for class

### DIFF
--- a/src/Objects/BaseObject.php
+++ b/src/Objects/BaseObject.php
@@ -60,6 +60,7 @@ abstract class BaseObject extends Collection
             return $relations[$property]::make($value);
         }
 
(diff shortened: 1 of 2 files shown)


jonnywilliamson:
http://php.net/manual/en/migration70.incompatible.php


```
Invalid class, interface and trait names ¶

The following names cannot be used to name classes, interfaces or traits:

bool
int
float
string
NULL
TRUE
FALSE
Furthermore, the following names should not be used. Although they will not generate an error in PHP 7.0, they are reserved for future use and should be considered deprecated.

resource
object
mixed
numeric
```

irazasyed:
Thanks!

PR #294, opened 2016-11-14
Improve Update->getMessage()

sloothword:
It does make more sense if `Update->getMessage()` returns any contained message.

Alternatively we could refactor it to `getIncludedMessage()`, as now $update->getMessage() and $update->message could return differently.


irazasyed:
Thanks, @sloothword for your contribution.

PR #326, opened 2017-01-18
editMessageText added

dena-a:


irazasyed:
Thanks for sending the PR. This method is already implemented in [develop](https://github.com/irazasyed/telegram-bot-sdk/blob/develop/src/Methods/EditMessageText.php) branch.

I'll be merging it soon after some testing. Thanks again 👍

PR #336, opened 2017-02-08
Add InlineButton Function

InSide24:
Hello would be great if you API will have a InlineButton option.

https://core.telegram.org/bots/api#inlinekeyboardmarkup

Kind Regards ;)


irazasyed:
It's already present @InSide24 

Please check this [file](https://github.com/irazasyed/telegram-bot-sdk/blob/develop/src/Keyboard/Keyboard.php).

PR #384, opened 2017-05-08
Proposed updates

jonnywilliamson:
Please do not merge, this PR is to discuss my proposed changes/updates.

Many small commits are made to make it easy to revert something.

### DIFF
--- a/src/Commands/CommandBus.php
+++ b/src/Commands/CommandBus.php
@@ -204,7 +204,7 @@ class CommandBus extends AnswerBus
     {
         $message = $update->getMessage();
 
-        if ($message->has('entities')) {
+        if (!is_null($message) && $message->has('entities')) {
             $this->parseCommandsIn($message)
                 ->each(function ($botCommand) use ($update) {
                     $this->process($botCommand, $update);
--- a/src/Commands/CommandBus.php
+++ b/src/Commands/CommandBus.php
@@ -204,12 +204,12 @@ class CommandBus extends AnswerBus
     {
         $message = $update->getMessage();
 
-        if ($message !== null && $message->has('text') && $message->has('entities')) {
-
+        if ($message->has('entities')) {
             $this->parseCommandsIn($message)
                 ->each(function ($botCommand) use ($update) {
                     $this->process($botCommand, $update);
                 });
+
         }
 
         return $update;
--- a/src/Keyboard/Keyboard.php
+++ b/src/Keyboard/Keyboard.php
@@ -129,7 +129,7 @@ class Keyboard extends Base
      *
      * @link https://core.telegram.org/bots/api#inlinekeyboardbutton
      *
-     * @param string|array $params                           [
+     * @param array $params                           [
      *
      * @type string        $text                             Required. Label text on the button
      * @type string        $url                              Optional. HTTP url to be opened when button is pressed
@@ -140,11 +140,11 @@ class Keyboard extends Base
      *
      * ]
      *
-     * @return string
+     * @return array
      */
-    public static function inlineButton($params = []): string
+    public static function inlineButton(array $params = [])
     {
-        return self::button($params);
+        return $params;
     }
 
     /**
--- a/src/Keyboard/Keyboard.php
+++ b/src/Keyboard/Keyboard.php
@@ -24,16 +24,13 @@ namespace Telegram\Bot\Keyboard;
  * ];
  * </code>
  *
- * @method $this setResizeKeyboard($boolean)     Optional. Requests clients to resize the keyboard vertically for
- *         optimal fit.
- * @method $this setOneTimeKeyboard($boolean)    Optional. Requests clients to hide the keyboard as soon as it's been
- *         used.
- * @method $this setSelective($boolean)          Optional. Use this parameter if you want to show the keyboard to
- *         specific users only.
+ * @method $this setResizeKeyboard($boolean)     Optional. Requests clients to resize the keyboard vertically for optimal fit.
+ * @method $this setOneTimeKeyboard($boolean)    Optional. Requests clients to hide the keyboard as soon as it's been used.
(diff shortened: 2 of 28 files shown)


src/Keyboard/Keyboard.php
```diff
@@ -147,21 +152,23 @@ public static function inlineButton($params = []): string
...
-     * @var bool    $params ['selective']
+     * @type bool   $remove_keyboard Required. Requests clients to remove the custom keyboard (user will not be able to summon this keyboard; if you want to hide the keyboard from sight but keep it accessible, use one_time_keyboard in ReplyKeyboardMarkup)
+     * @type bool   $selective       Optional. Use this parameter if you want to remove the keyboard for specific users only. Targets: 1) users that are @mentioned in the text of the Message object; 2) if the bot's message is a reply (has reply_to_message_id), sender of the original message.
+     *
+     * ]
      *
      * @return string
      */
-    public static function hide(array $params = []): string
+    public static function remove(array $params = []): string
```
irazasyed: Return would be `Keyboard` if it's keyboard for the next method too. See `forceReply`.

src/Keyboard/Keyboard.php
```diff
@@ -24,16 +24,13 @@
...
- * @method $this setSelective($boolean)          Optional. Use this parameter if you want to show the keyboard to
- *         specific users only.
+ * @method $this setResizeKeyboard($boolean)     Optional. Requests clients to resize the keyboard vertically for optimal fit.
+ * @method $this setOneTimeKeyboard($boolean)    Optional. Requests clients to hide the keyboard as soon as it's been used.
+ * @method $this setSelective($boolean)          Optional. Use this parameter if you want to show the keyboard to specific users only.
  */
 class Keyboard extends Base
 {
-    /** @var bool Make an Inline Keyboard */
+    /** @type bool Make an Inline Keyboard */
```
irazasyed: Why is it changed to `@type`?

src/Traits/Validator.php
```diff
@@ -12,14 +12,14 @@
...
      * Determine given param in params array is a file id.
      *
-     * @param $param
+     * @param $inputFileField
      * @param $params
      *
      * @return bool
      */
-    protected function hasFileId($param, array $params): bool
+    protected function hasFileId($inputFileField, array $params): bool
```
irazasyed: We should type hint this to string if possible.

src/Traits/Http.php
```diff
@@ -253,15 +267,17 @@ protected function prepareMultipartParams(array $params, $inputFileField): array
...
-
+        //All files should be provided in an InputFile object
         if (is_resource($inputFile)) {
             throw CouldNotUploadInputFile::resourceShouldBeInputFileEntity($inputFileField);
         }
 
+        //If the user provides a path to a file, attempt to create an InputFile Object automatically for them.
         if (is_string($inputFile)) {
             $params[$inputFileField] = InputFile::create($inputFile);
         }
```
irazasyed: Are we getting rid of these 2 checkers and instead just have one to check if it's an instance of InputFile?

src/Commands/CommandBus.php
```diff
@@ -204,12 +204,12 @@ protected function handler(Update $update): Update
     {
         $message = $update->getMessage();
 
-        if ($message !== null && $message->has('text') && $message->has('entities')) {
-
+        if ($message->has('entities')) {
```
irazasyed: What happens when the `$message` is `null` and there is no `has()` method?

src/Methods/Chat.php
```diff
@@ -73,9 +71,7 @@ public function kickChatMember(array $params): bool
      */
     public function leaveChat(array $params): bool
     {
-        $this->get('leaveChat', $params);
-
-        return true;
+        return $this->get('leaveChat', $params)->getResult();
```
irazasyed: Why is this method not cast to `bool` while the other methods have (See `kickChatmember` & `unbanChatMember`)?

Why is it even being cast, though?

src/Traits/Http.php
jonnywilliamson: But how do you detect if its a file_id string or a file path string?

I thought you said you wanted to allow file_id to still be passed as a string?

src/Methods/Chat.php
jonnywilliamson: Initially I thought the result was coming back as a string. I checked again and it is a bool so I've removed the cast on the others.. Good catch.

src/Commands/CommandBus.php
jonnywilliamson: Yeah, I wanted to know your thoughts on the `->getMessage()` method.

At the moment it returns a "message (collection)" or "edited_message (collection)" or null.

I think it might be appropriate to return an EMPTY collection() rather than null....but I wanted to discuss. In the mean time I've reverted to checking for null. Will push soon.

src/Keyboard/Keyboard.php
jonnywilliamson: Because my find and replace regex string for the others ones caught it by mistake!!

Reverted

src/Traits/Validator.php
jonnywilliamson: Done.

src/Traits/Http.php
irazasyed: `uploadFile` is already checking if it's a file_id and if yes, then does a post request directly. This is if it's not a file_id part. So if it's not a file_id, then it's either a URL or file path. In any case, we need to use InputFile entity to upload such files.

src/Methods/Chat.php
irazasyed: Cool. Thanks!

src/Commands/CommandBus.php
irazasyed: That totally makes sense! Returning empty collection. We need to do this to all other methods where it's simply null. Because it ends up failing all `has` and other methods.

irazasyed:
Thank you so much for all your contributions. Highly appreciated 💯👍

PR #385, opened 2017-05-10
Update order of text/settings in the telegram.php config.

jonnywilliamson:
Tweaked some wording too.

### DIFF
--- a/src/Laravel/config/telegram.php
+++ b/src/Laravel/config/telegram.php
@@ -1,57 +1,61 @@
 <?php
-
 return [
+    /*
+    |--------------------------------------------------------------------------
+    | Your Telegram Bots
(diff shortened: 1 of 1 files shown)


src/Laravel/config/telegram.php
```diff
@@ -1,58 +1,62 @@
...
-    |
-    */
-    'default' => 'common',
-
-    /*
-    |--------------------------------------------------------------------------
-    | Telegram Bots
+    | Your Telegram Bots
     |--------------------------------------------------------------------------
+    | You may use as multiple bots at once  using the manager class. Each bot
```
irazasyed: Please remove the extra space between `...at once` and `using the manager..`

PR #387, opened 2017-05-10
Remove nulls and replace with empty collection.

jonnywilliamson:
Return an empty collection for better chaining and usability.


jonnywilliamson:
Sigh.

Squashed commits keep re-appearing. We'll have to fix this work flow.

PR #388, opened 2017-05-10
Remove arguments from commands as the detection process has completely changed.

jonnywilliamson:


ddzobov:
Why arguments was removed? What replacement for that?

Amegatron:
1) I also wonder what is the new detection process
2) `src/Commands/Command.php::185` - arguments are still passed to `handle`

PR #390, opened 2017-05-11
Add entity trait to help work with entities in commands etc.

jonnywilliamson:
See below

### DIFF
--- a/src/Commands/Command.php
+++ b/src/Commands/Command.php
@@ -27,6 +27,9 @@ abstract class Command implements CommandInterface
     /** @var string The Telegram command description. */
     protected $description;
 
+    /** @var int|null For multiple commands in an update, the current command number. */
+    protected $entityNumber;
+
     /**
      * Get Command Name.
      *
@@ -88,10 +91,11 @@ abstract class Command implements CommandInterface
     /**
      * @inheritDoc
      */
-    public function make(Api $telegram, Update $update)
+    public function make(Api $telegram, Update $update, $entityNumber = null)
     {
         $this->telegram = $telegram;
         $this->update = $update;
+        $this->entityNumber = $entityNumber;
 
         return $this->handle();
     }
--- a/src/Commands/CommandBus.php
+++ b/src/Commands/CommandBus.php
@@ -206,8 +206,8 @@ class CommandBus extends AnswerBus
 
         if (!is_null($message) && $message->has('entities')) {
             $this->parseCommandsIn($message)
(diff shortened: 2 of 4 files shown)


jonnywilliamson:
Hi

So I've had a bit of an idea for a "feature" or more like a trait.

I'd like to tell you what I'm thinking and you can look at the code and tell me if it's a good idea, or if there's a better way to do this.

So with the new command processing system we have, we know how many entities are in a message, where they are (offset) and how long the command is (length). What we don't have is an easy way to get access to any "arguments" that a command might have.

This now gets more complicated because we can now have more than one command in a message, and even different types of entities between command entities like urls or formatting stuff.

So take a message like this

"Hi there bot, get me a /image cat and http://example.com and also a /image dog"

This message would have 3 entities,

0 => command
1 => url
2 => command

But say the image command needs an argument and for sake of discussion, we have decided it can only take one word for it. To get access to the argument, we need to know a few things.

a) Where is the command in the message?
b) Which command are we talking about the 0 or 2 one?

Getting the answer to (a) is easy, as the entity array tells us exactly where the command is and how long it is.
The answer to (b) is more complicated. When we process multiple commands, we currently have no idea which command is currently being processed.

So I added some code to pass along the key of the command we are working with during the process loop. That means that on the command class is a new property called `entityNumber`, if the command needs to know which "number of command" it is working with (ie in this case command 0 or 2) we can now access that value easily.


SO why would we want to know this?

Well, this leads me to the next part of the code.

I created a trait called 'Entity' which allows us to extract all the words/strings AFTER the location of the current entity being processed.

So in the example above, if we had an "image command class" the flow would look like this:

* First entity would be detected and the image command class would be executed.
* The image class asks for all the details of entity 0 from the update object
* It then extracts all the words/string after the first /image in the message text and gets an array (collection) back like this:

	[
		'cat'
		'and',
		'http://example.com',
		'and',
		'also',
		'a',
		'/image',
		'dog',
	]

The image class can decide how many/which words it wants to use. In our case probably ->first() would work.
* Then the next entity would be detected but its a url and not processed as a command.
* The second command entity would be detected and sent again to the Image command class.
* This time the command would ask for details on entity 2.
* Again it would get all the strings after the SECOND entity and return an array like this
	[
		'dog',
	]


So that was quite a long winded explanation of what I thought might be useful. I found it hard thinking of descriptive method names in the trait, but if you get a chance to look at the code, perhaps you might think of either a better way to do this or some better names.

Thoughts!?

jonnywilliamson:
Replaced with new command parsing system

PR #415, opened 2017-07-15
Sending files by url bug fixed.

dena-a:
On sending files by url, CouldNotUploadInputFile exception had been throws.
Detecting url added to uploadFile method.

### DIFF
--- a/src/Traits/Validator.php
+++ b/src/Traits/Validator.php
@@ -23,6 +23,19 @@ trait Validator
     }
 
     /**
+     * Determine given param in params array is a url.
+     *
+     * @param string $inputFileField
+     * @param array  $params
+     *
+     * @return bool
+     */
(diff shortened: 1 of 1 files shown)


irazasyed:
Thanks for the PR. Haven't tested this but can you please tell me if Telegram handles the image downloading if we send them the URL directly?

Last time when I checked, they had many restrictions and or it wasn't possible. We dropped url validation for some reason back then and I'm sure we had a reason for doing it.

jonnywilliamson:
Files sent via url must be inclosed with a InputFile class

```php
    \Telegram::bot('mybot')->sendPhoto([
        'chat_id'              => '12345',
        'photo'           => InputFile::create('http://mydomain.com/image.jpg', 'optionalImageName.jpg'),
    ]);

```

PR #418, opened 2017-07-24
pass through params from commandsHandler() to getUpdates()

Overtorment:
I'm using `$update = $telegram->commandsHandler();` in production and discovered that it doesn't use long polling. To make it use long polling I had to rewrite `commandsHandler()` with a call to `getUpdates()` with  `timeout` parameter set.

It would be much simpler to passthrough `timeout` from the scope I'm calling `commandsHandler()` in the first place.
I.e.  `$update = $telegram->commandsHandler(false, ['timeout' => 30]);`

### DIFF
--- a/src/Api.php
+++ b/src/Api.php
@@ -1309,11 +1309,12 @@ class Api
     /**
      * Processes Inbound Commands.
      *
-     * @param bool $webhook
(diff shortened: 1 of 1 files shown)


irazasyed:
Looks good. Thanks 👍

PR #430, opened 2017-08-19
Fixbug CommandBus $entity Invalid variable.

mylukin:
1. fixbug CommandBus $entity Invalid variable.
2. add bot name, use bots for asynchronous queue


irazasyed:
Thanks for the PR. Please create separate PRs for each major change. It's easier to review and merge that way than everything combined like this one. Thanks!

PR #437, opened 2017-09-13
Fix

Ishodnikov:
Why uploadFile? Must be post.


irazasyed:
Because Telegram also accepts SSL cert that you can send it to them. Please refer docs.

PR #438, opened 2017-09-13
Add User::getIsBot()

holtkamp:
https://core.telegram.org/bots/api#august-23-2017
https://core.telegram.org/bots/api#user


irazasyed:
You think you can also update this in develop branch? That has many changes and that's what is going to be released going forward. Please create a new PR if possible.

Thanks 👍

PR #444, opened 2017-09-17
flatten code

imanghafoori1:


irazasyed:
Thanks for the contributions. I think you need to look into the develop branch as well. Because that's what is going to be released going forward. Master is quite outdated for now. I'll merge this but please consider contributing to develop branch. Thanks!

PR #455, opened 2017-09-19
create artisan command

imanghafoori1:


src/Laravel/WebHookArtisanCommand.php
```diff
@@ -0,0 +1,67 @@
...
+        parent::__construct();
+    }
+
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'telegram:webhook
+                {--setup : To declare your webhook on telegram servers. So the can call you.}
```
jonnywilliamson: Typo 

So THEY can call you

jonnywilliamson:
Haven't checked this but it looks good. Nice job.

jonnywilliamson:
Also, looks like one of the tests is failing.

It's being passed an object instead of an array. Probably a simple fix. But would be good to have it fixed if you can. Thanks

jonnywilliamson:
@irazasyed is going to fall over when he sees the number of commits. It'll twitch at his OCD 😆 😀

imanghafoori1:
Previous commits are already merged.

imanghafoori1:
I do not know why is that failing...
I have created a new command class and added it in the service provider.
Is it introduced in the 2 recent commits?

irazasyed:
Please create a new PR with only the new addition/changes. You might wanna rebase your repo and also create a new branch if required.

PR #456, opened 2017-09-19
refactorings

imanghafoori1:


imanghafoori1:
I tried rebase ... but it did not work

PR #457, opened 2017-09-19
Artisan command

imanghafoori1:
Finally, I found the problem.

### DIFF
--- a/src/Laravel/TelegramServiceProvider.php
+++ b/src/Laravel/TelegramServiceProvider.php
@@ -48,7 +48,7 @@ class TelegramServiceProvider extends ServiceProvider
     {
         $this->registerManager($this->app);
         $this->registerBindings($this->app);
-        $this->commands('telegrambot.commands.webhook');
+        $this->commands('telegram.bot.commands.webhook');
     }
 
     /**
@@ -81,7 +81,7 @@ class TelegramServiceProvider extends ServiceProvider
             return $manager->bot();
         });
 
-        $this->app->singleton('telegrambot.commands.webhook', function () {
+        $this->app->singleton('telegram.bot.commands.webhook', function () {
             return new WebHookArtisanCommand();
         });
 
--- a/src/Laravel/TelegramServiceProvider.php
+++ b/src/Laravel/TelegramServiceProvider.php
@@ -8,6 +8,7 @@ use Illuminate\Support\ServiceProvider;
 use Illuminate\Contracts\Container\Container as Application;
 use Laravel\Lumen\Application as LumenApplication;
 use Illuminate\Foundation\Application as LaravelApplication;
+use Telegram\Bot\Laravel\Artisan\WebhookCommand;
 
 /**
  * Class TelegramServiceProvider.
@@ -82,7 +83,7 @@ class TelegramServiceProvider extends ServiceProvider
         });
 
         $this->app->singleton('telegram.bot.commands.webhook', function () {
-            return new WebHookArtisanCommand();
+            return new WebhookCommand();
         });
 
         $app->alias('telegram.bot', Api::class);
--- a/src/Laravel/TelegramServiceProvider.php
+++ b/src/Laravel/TelegramServiceProvider.php
@@ -48,6 +48,7 @@ class TelegramServiceProvider extends ServiceProvider
     {
         $this->registerManager($this->app);
         $this->registerBindings($this->app);
+        $this->commands('telegrambot.commands.webhook');
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/Laravel/TelegramServiceProvider.php
```diff
@@ -48,6 +48,7 @@ public function register()
     {
         $this->registerManager($this->app);
         $this->registerBindings($this->app);
+        $this->commands('telegrambot.commands.webhook');
```
irazasyed: I think we should change this to `telegram.bot.commands.webhook`.

src/Laravel/TelegramServiceProvider.php
```diff
@@ -80,6 +81,10 @@ protected function registerBindings(Application $app)
             return $manager->bot();
         });
 
+        $this->app->singleton('telegrambot.commands.webhook', function () {
```
irazasyed: Update this as well ofc.

src/Laravel/WebHookArtisanCommand.php
```diff
@@ -0,0 +1,67 @@
...
+class WebHookArtisanCommand extends LaravelGeneratorCommand
+{
+    /**
+     * Create a new command instance.
+     *
+     */
+    public function __construct()
+    {
+        parent::__construct();
+    }
```
irazasyed: Do we need the constructor for this command? I think you can omit it for now.

src/Laravel/TelegramServiceProvider.php
imanghafoori1: This is a string. The dots as you know have no technical meaning.
It is a matter of human test to put dots.
I put them.

src/Laravel/WebHookArtisanCommand.php
imanghafoori1: For injecting other classes. Like Api class i think...

imanghafoori1:
The last to commits are part of my tinkerings.
And have no effects

src/Laravel/WebHookArtisanCommand.php
```diff
@@ -41,7 +44,9 @@ public function __construct()
     public function handle()
     {
         $botName = $this->argument('botName');
-        $this->token = config("telegram.bots.$botName.token");
+        $token = config("telegram.bots.$botName.token");
+        $this->telegram = new Api($token);
+
```
irazasyed: Inject `BotsManager` and access the methods/config.

Example:
```php
use Telegram\Bot\BotsManager;

// ...
// ....

/** @var Telegram\Bot\Api */
protected $telegram;

/** @var array */
protected $config = [];

public function handle(BotsManager $telegram)
{
    $botName = $this->argument('botName');

    $this->telegram = $telegram->bot($botName);
    $this->config = $telegram->getBotConfig($botName);
}

private function setupWebhook()
{
    $webhookUrl = array_get($this->config, 'webhook_url');
    $certificatePath = array_get($this->config, 'certificate_path', false);

    $params = ['url' => $webhookUrl];
    
    if ($certificatePath) {
        $params['certificate'] = $certificatePath;
    }

    $response = $this->telegram->setWebhook($params);

    $this->info($response);
}
```

src/Laravel/config/telegram.php
```diff
@@ -34,6 +34,8 @@
         'mybot' => [
             'username' => 'TelegramBot',
             'token'    => env('TELEGRAM_BOT_TOKEN', 'YOUR-BOT-TOKEN'),
+            'certificatePath'    => env('TELEGRAM_CERTIFICATE_PATH', 'YOUR-CERTIFICATE-PATH'),
+            'webhookUrl'    => env('TELEGRAM_WEBHOOK_URL', 'YOUR-BOT-WEBHOOK-URL'),
```
irazasyed: Use snake case please! Ex: `webhook_url`

src/Laravel/WebHookArtisanCommand.php
```diff
@@ -0,0 +1,93 @@
+<?php
+
+namespace Telegram\Bot\Laravel;
+
+use Illuminate\Console\GeneratorCommand as LaravelGeneratorCommand;
+use Symfony\Component\Console\Input\InputOption;
+use Telegram\Bot\Api;
+use Telegram\Bot\BotsManager;
+
+class WebHookArtisanCommand extends LaravelGeneratorCommand
```
irazasyed: Please rename this to `WebhookCommand` (Notice **Webhook** Not **WebHook**). Obviously update everywhere.

Perhaps, we could also move this into its own namespace like `Telegram\Bot\Laravel\Artisan` as I think we can add some more useful commands in near future. Thanks!

irazasyed:
Thanks 👍

PR #470, opened 2017-10-09
First attempt at new command argument parsing strategy.

jonnywilliamson:
New command argument parsing strategy. With tests. For review.


jonnywilliamson:
Also I think we should ditch hhvm for travis

irazasyed:
@jonnywilliamson Makes sense about HHVM. Let's do it.

Also, Thanks for the PR.

PR #475, opened 2017-10-12
New official features

jonnywilliamson:
All the following API changes have been implemented.

Not everything could be tested, due to me not having a payment system setup etc. 

Might be better not to squash all these commits this time.

image

image


irazasyed:
Thanks, much needed 👍 

And yeah, we'll have to look into the payment API.

PR #500, opened 2017-11-10
Update Api.php for deleting message

mosleim:
Adding Delete Message Function


irazasyed:
It's already implemented in `develop` branch. Please use that.

PR #525, opened 2018-02-01
\CommandsHandler::updatesTruncate

iqdoctor:
Truncate inbound updates Queue on Telegram server in new method \Telegram\Bot\Traits\CommandsHandler::updatesTruncate

I'd like to suggest feature CommandsHandler::updatesTruncate.

Rationale: if an exception is thrown in the body of the command handler, updates (including those containing commands) are not deleted from the server, and the next attempt to request and process messages from the server will also end with an error in the same place.

After the interception (in the main loop) of the unhandled exception inside the command itself, you need to imagine the ability to clear (truncate) the command queue on the server.

### DIFF
--- a/src/Commands/Command.php
+++ b/src/Commands/Command.php
@@ -317,7 +317,7 @@ abstract class Command implements CommandInterface
 
     private function cutTextBetween(Collection $splice)
     {
-        return substr(
+        return mb_substr(
             $this->getUpdate()->getMessage()->text,
             $splice->first(),
             $splice->last() - $splice->first()
@@ -326,7 +326,7 @@ abstract class Command implements CommandInterface
 
     private function cutTextFrom(Collection $splice)
     {
-        return substr(
+        return mb_substr(
             $this->getUpdate()->getMessage()->text,
             $splice->first()
         );
--- a/src/Commands/CommandBus.php
+++ b/src/Commands/CommandBus.php
@@ -5,6 +5,7 @@ namespace Telegram\Bot\Commands;
 use Telegram\Bot\Api;
 use Telegram\Bot\Objects\Update;
 use Illuminate\Support\Collection;
+use Telegram\Bot\Objects\User;
 use Telegram\Bot\Traits\Singleton;
 use Telegram\Bot\Answers\AnswerBus;
 use Telegram\Bot\Exceptions\TelegramSDKException;
@@ -15,6 +16,11 @@ use Telegram\Bot\Exceptions\TelegramSDKException;
 class CommandBus extends AnswerBus
 {
     use Singleton;
+
+    /**
+     * @var User
+     */
+    protected $me;
 
     /**
      * @var Command[] Holds all commands.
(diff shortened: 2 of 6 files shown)


irazasyed:
Thanks for opening the PR. While this makes sense, I'm not sure why we need to get the update here?

It's as simple as responding Telegram server's with a status code 200 while logging the error.

What you can do is, catch the exception, log it internally and just respond with 200 status (Echo something to the body, that's all). So that update will automatically get marked as read (This is in case of a webhook update).

If I got this wrong, please let me know an example of how you plan to use this method?

Thanks!

iqdoctor:
Thanks for the answer and for creating the diamond library :-), Said.
The main purpose of CommandsHandler::updatesTruncate() method is to use it in conjunction with the Update::getUpdates() method.
My bot is running behind a firewall that does not allow the use of web hooks. 
More generally, batch receipt (via getUdates()) and processing of all updates for one call seems profitable to me.
In this case the Command::handle() method will be called by lib for each command in update queue. If this method in my command class throws an uncaught exception, the processing of command queue is interrupted. 
In this case, I would like to clear the entire queue on the server and continue.
It is clear that a good style is to intercept all exceptions in Command :: handle (), but you can not completely exclude throwing an exception from it. But I would like to have a reliable code with recovery after any types of failures.
Below is an example from my current code.

`private function messageLoop(): void
{
   $this->state = self::STATE_RUNNING;
   while ($this->state === self::STATE_RUNNING) {
       try {
           $this->messagesAndCommandsHandler();
       } catch (\Exception $e) {
           echo $e->getMessage() . " " . $e->getTraceAsString();
           $this->telegram->updatesTruncate();
       }

   }
}

/**
* @throws AccessDeniedException
*/
private function messagesAndCommandsHandler(): void
{
   $updates = $this->telegram->getUpdates(['timeout' => 10]);

   $this->telegram->commandsHandler();
   …

   //цикл для ликвидации дребезга в нажатии кнопок на клавиатуре, воспринимаем только последнюю команду.
   foreach ($updates as $i) {
     //...
   }
}
`

src/Commands/CommandBus.php
```diff
@@ -205,13 +213,22 @@ protected function parseCommandsIn(Collection $message): Collection
...
+            $this->execute($command, $update, $entity);
+        }
+    }
+    
+    protected function isMyCommand($command, $bot) 
+    {
+        if ($this->me === null) {
+            $this->me =  $this->telegram->getMe();
+        }
+        return $bot === null || $this->me->username === $bot;
```
irazasyed: I think instead of making a call to `getMe()` API (Which would be too many calls for every command), we could just use the username set in the config file itself.

src/Commands/Command.php
```diff
@@ -326,7 +326,7 @@ private function cutTextBetween(Collection $splice)
 
     private function cutTextFrom(Collection $splice)
     {
-        return substr(
+        return mb_substr(
```
irazasyed: Please create a separate PR for this.

PR #526, opened 2018-02-07
Added support of the WebhookInfo method

zagovorichev:


zagovorichev:
@irazasyed I have no idea what does it means with all these errors. Do I need to fix them? I haven't done so many changes to be responsible for all CI issues :)

irazasyed:
Hi @zagovorichev, Thanks for the PR. This method already exists in our `dev-develop` branch, [here](https://github.com/irazasyed/telegram-bot-sdk/blob/develop/src/Methods/Update.php#L114-L127).

Please use that branch (Note: Major changes between the two which are yet to be documented).

PR #536, opened 2018-03-08
Added @method to Telegram Facade

roelofr:
Hey,

I've noticed my IDE (phpStorm) is having trouble figuring out how a Facade is linked to an object, causing the autocomplete to belch.

This commit solves it, but it's not the best thing ever, since it involves adding all methods to the phpdoc of the Telegram facade.

Maybe this could be automated, using the Reflection package, but I'll have to look into that.

---
PS: Your contribution guide discusses a `CHANGELOG.md`, but I can't find any :confused:


irazasyed:
Use the [IDE Helper](https://github.com/barryvdh/laravel-ide-helper).

roelofr:
Oh, didn't know about that. Thanks!

PR #549, opened 2018-03-28
Version 2 - Сallback query

2at2:
Callback query supporting


2at2:
Does travis and stylecli pass on 2.0 branch?

PR #569, opened 2018-04-25
[develop] Pass command classname for triggering

ddzobov:
Example: $this->triggerCommand(HelpCommand::class);


irazasyed:
This is good. Thanks!

PR #596, opened 2018-10-09
BotAPI4.0

jonnywilliamson:
Update API to latest version, however Telegram Passport is stil WIP

### DIFF
--- a/src/Api.php
+++ b/src/Api.php
@@ -27,6 +27,7 @@ class Api
         Methods\Get,
         Methods\Location,
         Methods\Message,
+        Methods\Passport,
         Methods\Payments,
         Methods\Query,
         Methods\Stickers,
--- a/src/Methods/EditMessage.php
+++ b/src/Methods/EditMessage.php
@@ -2,6 +2,7 @@
 
 namespace Telegram\Bot\Methods;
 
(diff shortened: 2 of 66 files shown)


irazasyed:
Hey @jonnywilliamson !!

I'm very well, thanks for asking! Hope you're good too, Long time bud :)

And thanks for the PR as always highly appreciated. Been away and couldn't catch up with these changes myself. I'll look review this soon and merge if everything is good.

As far as Passport is concerned, I haven't had any use case either. So didn't check but will hopefully check soon and we can discuss on it later. Thanks!

irazasyed:
I think we can bump our min. required version for PHP to 7.2 now as 7.3 is already right around the corner and has some nice stuff coming up. I wouldn't mind going straight to the 7.3 either but I think that would be too early for others to upgrade.

And with the 7.2 requirements, we can optimize the code further and also use the collection pipeline/other methods (if you remember we had to get rid of it back then). And further typehint the returns too. Let me know your thoughts. 

Others are also most welcome to add their inputs as well :)

jonnywilliamson:
I see you have merged! 😨 !!

Well I should be around over next month or 2 to tidy it up. Maybe we'll find a convenient time and chat about it soon.

Take care!

PR #600, opened 2018-10-24
Use laravel request content

GrKamil:


irazasyed:
Thanks for the PR. The library is not limited to Laravel usage and can be used outside of Laravel framework as well.

PR #613, opened 2018-12-06
Append proxy support

finagin:
Support proxy by the environment variable `TELEGRAM_BASE_URL`


jonnywilliamson:
Hi @finagin 

I actually don't think using env() here is a good idea. This package gets used alot with laravel, (not exclusively though).

The docs state this

https://laravel.com/docs/5.7/configuration#configuration-caching
image


So having this in this package means that a person who implements caching the config will always get NULL returned when this is run.

PR #659, opened 2019-08-13
Bot API 4.4

jonnywilliamson:
July 29, 2019

Added support for animated stickers. New field is_animated in Sticker and StickerSet objects, animated stickers can now be used in sendSticker and InlineQueryResultCachedSticker.

Added support for default permissions in groups. New object ChatPermissions, containing actions which a member can take in a chat. New field permissions in the Chat object; new method setChatPermissions.

The field all_members_are_administrators has been removed from the documentation for the Chat object. The field is still returned in the object for backward compatibility, but new bots should use the permissions field instead.

Added support for more permissions for group and supergroup members: added the new field can_send_polls to ChatMember object, added can_change_info, can_invite_users, can_pin_messages in ChatMember object for restricted users (previously available only for administrators).

The method restrictChatMember now takes the new user permissions in a single argument of the type ChatPermissions. The old way of passing parameters will keep working for a while for backward compatibility.

Added description support for basic groups (previously available in supergroups and channel chats). You can pass a group's chat_id to setChatDescription and receive the group's description in the Chat object in the response to getChat method.

Added invite_link support for basic groups (previously available in supergroups and channel chats). 

You can pass a group's chat_id to exportChatInviteLink and receive the group's invite link in the Chat object in the response to getChat method.

File identifiers from the ChatPhoto object are now invalidated and can no longer be used whenever the photo is changed.

All webhook requests from the Bot API are now coming from the subnets 149.154.160.0/20 and 91.108.4.0/22. Most users won't need to do anything to continue receiving webhooks. If you control inbound access with a firewall, you may need to update your configuration. You can always find the list of actual IP addresses of servers used to send webhooks there: https://core.telegram.org/bots/webhooks.

As of the next Bot API update (version 4.5), nested MessageEntity objects will be allowed in message texts and captions. Please make sure that your code can correctly handle such entities.

### DIFF
--- a/src/Methods/Chat.php
+++ b/src/Methods/Chat.php
@@ -5,6 +5,7 @@ namespace Telegram\Bot\Methods;
 use Telegram\Bot\Traits\Http;
 use Telegram\Bot\Objects\ChatMember;
 use Telegram\Bot\FileUpload\InputFile;
+use Telegram\Bot\Objects\ChatPermissions;
 use Telegram\Bot\Objects\Chat as ChatObject;
 use Telegram\Bot\Exceptions\TelegramSDKException;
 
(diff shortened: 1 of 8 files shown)


irazasyed:
Thank you 👍

PR #664, opened 2019-09-06
laravel helper function change

eric2100:
array_get function change Arr::get
str_contains change Str::contains
snake_case change Str::snake


irazasyed:
Merged in #665

PR #665, opened 2019-09-07
Fixed removed string/array helpers for laravel 6

nickurt:
String/Array Helpers have been removed in Laravel 6.0, (see https://laravel.com/docs/6.0/upgrade#helpers), fixed it to use the FQCN.


nickurt:
Can you tag a new version in the 2.x range aswell? @irazasyed

irazasyed:
@nickurt That would require the merging of `master` branch which I think will break a few things due to many changes. I don't use 2.0. Perhaps you could send another PR with these changes to 2.0 branch?

PR #674, opened 2019-10-04
Added FileInputTest

omentes:
For issue #5

### DIFF
--- a/tests/FileInputTest.php
+++ b/tests/FileInputTest.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace Telegram\Bot\Tests;
+
+use Telegram\Bot\FileUpload\InputFile;
+
+class FileInputTest extends \PHPUnit_Framework_TestCase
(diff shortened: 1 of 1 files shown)


jonnywilliamson:
Wouldn't this fail if I ran the tests offline?

I work 85% without an internet connection when programming. It sucks.

omentes:
> Wouldn't this fail if I ran the tests offline?
> 
> I work 85% without an internet connection when programming. It sucks.

@jonnywilliamson I think I fixed it

irazasyed:
Hey @omentes, Thanks for the PR. Could you please write tests for the `develop` branch instead and send a PR there. Master will most likely retire in some time. Develop has become the primary as of now (3.0) and I'll be changing the structure a bit later to make things easier and less confusing. Too many changes in both branches have already caused a lot of conflicts unfortunately and we can't merge both. Thanks!

PR #688, opened 2019-10-31
studly_case -> Str::studly

autowp:


Sergo96:
@irazasyed hey, I'm wondering if this change is available when downloading package via composer? 

I installed ^2.0 version in my composer.json (2.3.0 locked in composer.lock) but when I open Commands/Command.php there is still studly_case(...) instead of Str::studly(...)

Am I doing something wrong?

Thanks in advance!

richard4339:
@Sergo96 The fix was done in November, but the last release was in September. Hopefully they'll do a 2.3.1 release soon.

PR #691, opened 2019-11-13
fix issue in sendMediaGroup

alr2413:
In the "sendMediaGroup" method the "inputFileField" key has been incorrectly set as "video_note" instead of "media". Also, since the "media" field is JSON serialized string and not an input field, as mentioned in the Telegram API Docs, I updated the "validateInputFileField" in "Traits/Http.php" class to support this case.
Thanks.

### DIFF
--- a/src/Traits/Http.php
+++ b/src/Traits/Http.php
@@ -342,7 +342,7 @@ trait Http
         }
 
         //All file-paths, urls, or file resources should be provided by using the InputFile object
-        if ((!$params[$inputFileField] instanceof InputFile) && (!is_json($params[$inputFileField]))) {
+        if ((!$params[$inputFileField] instanceof InputFile) && (!$this->is_json($params[$inputFileField]))) {
             throw CouldNotUploadInputFile::inputFileParameterShouldBeInputFileEntity($inputFileField);
         }
     }
--- a/src/Traits/Http.php
+++ b/src/Traits/Http.php
@@ -342,7 +342,7 @@ trait Http
         }
 
         //All file-paths, urls, or file resources should be provided by using the InputFile object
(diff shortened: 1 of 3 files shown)


src/Traits/Http.php
```diff
@@ -342,7 +342,7 @@ protected function validateInputFileField(array $params, $inputFileField)
         }
 
         //All file-paths, urls, or file resources should be provided by using the InputFile object
-        if (! $params[$inputFileField] instanceof InputFile) {
+        if ((!$params[$inputFileField] instanceof InputFile) && (!is_json($params[$inputFileField]))) {
```
irazasyed: PHP doesn't have `is_json` method.

alr2413: Yes, I checked my code again and found that it is a function from another library with the following code.

```
function is_json($string)
{
	json_decode($string);
	return (json_last_error() == JSON_ERROR_NONE);
}
```

alr2413:
Please Reply my question:

Yes, I checked my code and found that it is a function from another library with the following code. Where should I place this function?

```
function is_json($string)
{
	json_decode($string);
	return (json_last_error() == JSON_ERROR_NONE);
}
```

_Originally posted by @alr2413 in https://github.com/irazasyed/telegram-bot-sdk/pull/691_

src/Traits/Http.php
irazasyed: Please add it to the [Validator](https://github.com/irazasyed/telegram-bot-sdk/blob/develop/src/Traits/Validator.php) trait.

alr2413: Thanks.
I've added this change.
Please check it again.

irazasyed:
Thanks!

Issue #697, opened 2019-12-26
where is deleteMessage method?


harmakit:
i've just updated laravel to 6.x and your package has updated too
but now there is no deleteMessage now :(

PR #698, opened 2019-12-26
add deleteMessage method

harmakit:
in addition to #697


jonnywilliamson:
Just for your info this is alreadh in develop branch (v3)


https://github.com/irazasyed/telegram-bot-sdk/blob/develop/src/Methods/EditMessage.php#L199-L235

harmakit:
got it

PR #705, opened 2020-01-19
Master

Ev1ch:
Broken replyKeyboardMarkup.
Fix src/Api.php file by adding:
use Telegram\Bot\Keyboard\Keyboard;
public function replyKeyboardMarkup(array $params) { return Keyboard::make($params); }
in it.


jonnywilliamson:
I think you might have made a mistake in your PR here. DId you apply this to the wrong branch?

Might be best to close this and try again.

PR #707, opened 2020-02-17
Update .travis.yml

sergiy-petrov:
add php 7.3 and 7.4
removed hhvm from allowed failures


sergiy-petrov:
Also, I've remove 7.1 and 7.2 versions of php from travis

irazasyed:
Thanks!

PR #715, opened 2020-03-03
Laravel 7 Support

matthewnessworthy:

PR #718, opened 2020-03-05
Laravel 7 Support

matthewnessworthy:
Same as https://github.com/irazasyed/telegram-bot-sdk/pull/715 , but targeting https://github.com/irazasyed/telegram-bot-sdk/tree/master

I'm not sure what the differences between the branches `master`, `develop`, and `2.0` are, hence the duplicate PR

Also, i think we need some tags created after this

@irazasyed perhaps you could clarify

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
   "require": {
     "php": ">=5.6.0",
     "guzzlehttp/guzzle": "~6.0",
-    "illuminate/support": "^5.0 || ^6.0",
+    "illuminate/support": "^5.0 || ^6.0 || ^7.0",
     "league/event": "^2.1"
   },
   "require-dev": {


irazasyed:
- `2.0`: This branch was supposed to hold the last stable 2.0 version with no major changes but unfortunately due to the number of users that are still using, I'm supporting it but in near future, this will be locked.

- `master`: This was supposed to be the latest version, i.e. `3.0` while the `develop` was supposed to be anything that's major, experimental and appropriate for the next version.

But due to 3.0 not properly being documented, the branches have currently caused confusion. I plan to eliminate one of these branches and stick to master and tags only with dedicated branches for version archives as soon as we move to another major version in future.

I'll tag and release once I commit some latest minor changes to the 2.0 and develop branch. Thanks!

matthewnessworthy:
Ah I see, makes sense

PR #721, opened 2020-03-12
Add proxy to GuzzleHttpClient

vokamut:


src/HttpClients/GuzzleHttpClient.php
```diff
@@ -109,6 +109,7 @@ private function getOptions(
         array $headers,
         $body,
         $options,
+        $proxy = null,
```
irazasyed: `$proxy` should go at the end.

vokamut: Fixed

irazasyed:
Thanks!

PR #730, opened 2020-04-11
Bot API 4.7

jonnywilliamson:
March 30, 2020

Added the method sendDice for sending a dice message, which will have a random value from 1 to 6. (Yes, we're aware of the “proper” singular of die. But it's awkward, and we decided to help it change. One dice at a time!)
Added the field dice to the Message object.
Added the method getMyCommands for getting the current list of the bot's commands.
Added the method setMyCommands for changing the list of the bot's commands through the Bot API instead of @BotFather.
Added the ability to create animated sticker sets by specifying the parameter tgs_sticker instead of png_sticker in the method createNewStickerSet.
Added the ability to add animated stickers to sets created by the bot by specifying the parameter tgs_sticker instead of png_sticker in the method addStickerToSet.
Added the field thumb to the StickerSet object.
Added the ability to change thumbnails of sticker sets created by the bot using the method setStickerSetThumb.

Please note. Testing the stickerset thumbnails was not done. Don't have the time to be making stickers!

### DIFF
--- a/src/Api.php
+++ b/src/Api.php
@@ -23,6 +23,7 @@ class Api
         Traits\CommandsHandler,
         Traits\HasContainer;
     use Methods\Chat,
+        Methods\Commands,
         Methods\EditMessage,
         Methods\Game,
         Methods\Get,
--- a/src/Methods/Commands.php
+++ b/src/Methods/Commands.php
@@ -1,0 +1,62 @@
+<?php
+
+namespace Telegram\Bot\Methods;
+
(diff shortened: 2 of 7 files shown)


irazasyed:
Thank you @jonnywilliamson :)

jonnywilliamson:
No problem!

Have found a very annoying issue with the singleton pattern in the command bus being used with laravel horizon.

Might like to chat with you about it over next while if you aren't busy?

Hope you're also keeping safe/well. 👌

Issue #649, opened 2019-05-17
triggerCommand failed on v3.0.0


smskin:
Laravel v5.8.17.
telegram-bot-sdk v.3.0.0

Method triggerCommand('help',$update) return exception "Call to a member function filter() on null".
![image](https://user-images.githubusercontent.com/3227797/57952794-e7566e80-78f6-11e9-825a-de80e17af5c7.png)
![image](https://user-images.githubusercontent.com/3227797/57952809-f2110380-78f6-11e9-898f-b2c9ad8f398f.png)

$update not empty. 
![image](https://user-images.githubusercontent.com/3227797/57952650-829b1400-78f6-11e9-9669-7ac13fc9f0e7.png)

romikzo:
+1 same problem

frasagal:
same problem!!

tanogalego:
Still happening :/

irazasyed:
Fixed #733

Issue #520, opened 2018-01-18
send Extra arguments to triggerCommand in laravel telegram bot


ahmadbadpey:
I'm using [irazasyed/telegram-bot-sdk/][1] laravel package to make a telegram bot for a simple shopping. 

I used inline queries to show many product until users can pick what those want. 

to show a list of products I wrote this : 

    public function Webhook ()
    {
    	$update = Telegram::commandsHandler(true);
    
    
    if ($update->isType('inline_query')) {
    				$inlineQuery   = $update->getInlineQuery();
    				$inlineQueryId = $inlineQuery->getId();
    
    				$query = $inlineQuery->getQuery();
    
    				$category    = Category::where('cat_alias', $query)->first();
    				$catProducts = $category->products;
    
    				if (!empty($catProducts)) {
    					$results = array ();
    
    					foreach ($catProducts as $product) {
    						$results[] = array (
    							'type'         => 'article',
    							'id'           => $product->product_id,
    							'title'        =>  $product->title
    						);
    					}
    
    
    					Telegram::answerInlineQuery([
    						'inline_query_id' => $inlineQueryId,
    						'cache_time'      => 0,
    						'results'         => $results
    					]);
    
    
    				}
    
    
    			}
    else if ($update->isType('chosen_inline_result')) {
    
    
    				$chosenInlineResult   = $update->getChosenInlineResult();
    				$chosenInlineResultId = $product_id = $chosenInlineResult->getResultId();
    
    
    				Telegram::triggerCommand('show_product');
    
    			}
    
    }

As you can see in the second `else if ` I checked whether **update message** is type of `chosen_inline_result` or not? if it is , I fetch the **ID** of selected product and called another defined Command named `show_product` .

But whenever it called I got this error:

    Type error: Argument 2 passed to Telegram\Bot\Api::triggerCommand() must be an instance of Telegram\Bot\Objects\Update, none given

Even when send `$update` parameter to `triggerCommand` as second paramter I got this error in this case:

     Call to a member function getText() on null {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0):

I want in fact, when a user chosen an option, get the product ID and send it to `show_product` defined and manage other jobs there. 

  [1]: https://github.com/irazasyed/telegram-bot-sdk/

Issue #401, opened 2017-05-30
triggerCommand() not working


Devrandomuser:
This is the code of my callback handler in laravel:

```
Route::post('/telegram_webhook/b37fb36g', function(Request $request){

	 //will cover all "main" menue entries where we can use /command
	 $update = Telegram::commandsHandler(true);

	 switch ($update['message']['text']) {
	 	case 'lala':
	 			Telegram::triggerCommand('showMenue');
	 		break;
	 	
	 	default:

	 		Telegram::triggerCommand('showMenue');

			Telegram::sendMessage([
            'chat_id' => $update['message']['chat']['id'], 
            'text' => "Sorry, command <b>" . $update['message']['text'] . "</b> wasn't found :/",
            'parse_mode' => 'HTML',
        ]);
	 }
});
```

For some reason, when for example the text "lala" is received, the bot doesn't redirect to the shoeMenue Command. The same problem occured with the default case. The message under the showMenue in the default block is working fine. Any idea why the triggerCommand doesn't work? Of course, the name "showMenue" is correct.

irazasyed:
Please post your config and "showMenue" codes to verify.

ja-heller:
@irazasyed 
Of course, thanks for your fast reply, here we go.

callback handler (I'm using laravel latest version):

```
Route::post('/telegram_webhook/b37fb36g', function(Request $request){

	 //will cover all "main" menue entries where we can use /command
	 $update = Telegram::commandsHandler(true);

	 switch ($update['message']['text']) {
	 	case 'xyz':
	 			Telegram::triggerCommand('showMenue');
	 		break;

	 	default:

        Telegram::sendMessage([
            'chat_id' => $update['message']['chat']['id'], 
            'text' => "Sorry, command <b>" . $update['message']['text'] . "</b> wasn't found :/",
            'parse_mode' => 'HTML',
        ]);

        Telegram::triggerCommand('showMenue');

	 		break;
	 }
});
```

my "shoeMenue.php"

```
<?php

namespace App\TelegramCommands;

use Telegram;
use Telegram\Bot\Actions;
use Telegram\Bot\Commands\Command;
use App\User;

class MenueCommand extends Command
{
    protected $name = "showMenue";

    protected $description = "Start Command to get you started";
    
    public function handle($arguments)
    {
        //get last messsage, so we can get the chat_id
        $message_object = $this->getUpdate();

        $response = Telegram::sendMessage([
[...]

PR #733, opened 2020-04-18
Fix triggerCommand issues

### LINKED ISSUE #724: triggerCommand with arguments
In v3 (dev-develop), How I can trigger a command inside handle function of other command passing extra parameters?

for example 

```php
$this->triggerCommand('$delete',"delete command"); // as v2
```

jonnywilliamson:
Fixes #724 #649 #401 #520

### DIFF
--- a/src/Commands/Command.php
+++ b/src/Commands/Command.php
@@ -306,6 +306,10 @@ abstract class Command implements CommandInterface
         //Get all the bot_command offsets in the Update object
         $commandOffsets = $this->allCommandOffsets();
 
+        if ($commandOffsets->count() === 0) {
+            return $this->getUpdate()->getMessage()->text;
+        }
+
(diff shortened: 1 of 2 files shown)


irazasyed:
Thanks 👍

MrFiregore:
I try to use the Trait "CommandHandler" in my command classto send parameters when triggered other command and get the following error:

`PHP Fatal error: Access level to Telegram\Bot\Traits\CommandsHandler::getCommandBus() must be public (as in class Telegram\Bot\Commands\Command) in /E:/wamp64/www/telegramnew/src/Commands/StartCommand.php on line 9`

image

**SOLVED, BUT......**


I solved calling to:

`         $this->getTelegram()->triggerCommand('getcommands',$update);`
but, how I can pass arguments to the function triggered?

MrFiregore:
With this solution the only way that I find is rewritte the message of the update and then send to the triggerCommand...

image
 
image

PR #763, opened 2020-07-09
Replace "array_wrap" with "Arr::wrap"

andriihorpenko:
As this helper no longer exists in Laravel 7, it should be replaced with the `Arr::wrap` variant. I **strongly recommend** releasing a new version as this issue prevents developers from using Bot Payment API.

### DIFF
--- a/src/Methods/Payments.php
+++ b/src/Methods/Payments.php
@@ -7,6 +7,7 @@ use Telegram\Bot\Objects\Message;
 use Telegram\Bot\Objects\Payments\LabeledPrice;
 use Telegram\Bot\Objects\Payments\ShippingOption;
 use Telegram\Bot\Traits\Http;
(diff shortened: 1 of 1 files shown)


jonnywilliamson:
Do you use this SDK's code for your bot payments and invoices etc?

If so we would like to hear from you more. We'll merge that PR in shortly, when @irazasyed has a chance. Due to the current climate and lockdowns there's been a bit of delay in this project.

andriihorpenko:
Nice to hear!
Yes, I’m going to use this SDK to process payments via Telegram. This is what the flow looks like as for now:
User creates an order -> backend sends an invoice to the user on Telegram using Bot Payment API -> user pays for the goods -> changes reflect on the frontend.

irazasyed:
Thanks!

PR #772, opened 2020-08-12
Offset issue when triggering a command

KielD-01:
- Fixed issue with the offset, when no command has passed

### DIFF
--- a/src/Commands/Command.php
+++ b/src/Commands/Command.php
@@ -169,9 +169,9 @@ abstract class Command implements CommandInterface
     /**
      * Process Inbound Command.
      *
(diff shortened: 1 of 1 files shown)


dura0ok:
Thank you very much) These changes helped solve my problem ... When I created a team by hand and tried to make a trigger inside, there were problems
The only thing that can be triggered only from the client is a telegram, and not through the Command.php method
It's a little sad, but sooooo cool

KielD-01:
@irazasyed any chance to get this merged?

irazasyed:
Thanks for your patience and PR!

PR #778, opened 2020-08-22
Replace "array_intersect_key" with "intersectByKeys"

slavkluev:
This method is available since 5.5 version of illuminate/support.


irazasyed:
Thank you!

jonnywilliamson:
Completely forgot about that! Thanks for refactoring that! @slavkluev

PR #804, opened 2020-11-05
Add possible set custom base bot api url

KartaviK:
Due to Bot api telegram v5.0 we can run ow telegram bot server even on local machine


irazasyed:
Thanks for your contribution but this is already implemented in v4.x which is due for a release. Closing this for now.

CrazyBoy49z:
@irazasyed When is version 4 scheduled for release?

PR #813, opened 2020-11-26
Add php8 support

KartaviK:


irazasyed:
Thanks for opening the PR. I've merged #818 PR as its much cleaner solution but I appreciate your contribution nonetheless! 👍🏻

PR #818, opened 2020-12-05
Added PHP 8.0 support

roelofr:
Added PHP 8.0 support to Composer and GitHub Actions matrix

Some logic needed to be changed:

1. Singletons had `private function __clone` and `private function __wakeup`. I replaced these to throw `LogicExceptions` ("Errors that should lead to a fix in your code")
2. `call_user_func_array` was changed in PHP 8.0 to accept an associative array as named arguments. Wrapping them in `array_values` solves this issue.
3. Added `--ignore-platform-req=php` to GitHub actions, to allow testing old code on new PHP versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,11 +21,11 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [7.2, 7.3, 7.4]
+        php: ['7.2', '7.3', '7.4', '8.0']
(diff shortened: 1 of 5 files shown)


roelofr:
I just noticed I named my branch a bit dumb, I can re-open the PR on a differently named branch, but the commits should explain for themselves.

roelofr:
Rebased onto the new `develop`.

irazasyed:
Thank you 👍

PR #821, opened 2020-12-10
Fixed commands receiving argument names instead of values

roelofr:
As reported [here](https://github.com/irazasyed/telegram-bot-sdk/commit/84a3574aad6c677768b0d3354e054d863ef80f62#commitcomment-44974105).


irazasyed:
Thanks!

PR #829, opened 2021-01-02
Fix an error in validateInputFileField

mihard:
Found a logical error while tried to send image URLs with `sendMediaGroup` 
Also, maybe a good idea to move `is_string` into `is_json`


irazasyed:
Thanks!

PR #832, opened 2021-01-15
Anonymous Admins

alexmorbo:
From BotApi:
> Added the field sender_chat to the class Message, containing the sender of a message which is a chat (group or channel).

November 4, 2020
Introducing Bot API 5.0


irazasyed:
Thanks

PR #841, opened 2021-02-01
Update Http.php, Because of array to string conversion in Laravel 8

AmooEbrahim:
Because of an error that I occurred with ```(string)array``` in Laravel 8, I changed this line to ```json_encode``` to solve this problem.

### DIFF
--- a/src/Traits/Http.php
+++ b/src/Traits/Http.php
@@ -204,6 +204,10 @@ trait Http
      * @param array $params
      *
      * @return array
+     *
+     * If your laravel version is below 7
+     *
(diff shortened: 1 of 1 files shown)


amirhs712:
I'm having the same trouble, please merge this pr.

ramilS:
this works fine  
```
$keyboard = Keyboard::make()
                ->inline()
                ->row(
                    Keyboard::inlineButton([
                        'text' => 'Test Btn',
                        'callback_data' => 'callback_from_testbtn'
                    ])
                );
```

irazasyed:
What is the error? You're supposed to use the Keyboard class as shown in the above example and the string typecast will automatically return the json encoded string anyway.

PR #854, opened 2021-04-07
Fixed: method ReflectionParameter::getClass() is deprecated, PHP8

semsphy:
If the `PHP version >= 8.0.0` then `getType()` method will be used else keep the previous method.

### DIFF
--- a/src/Answers/AnswerBus.php
+++ b/src/Answers/AnswerBus.php
@@ -57,7 +57,12 @@ abstract class AnswerBus
         $container = $this->telegram->getContainer();
         $dependencies = [];
         foreach ($params as $param) {
-            $dependencies[] = $container->make($param->getClass()->name);
+            if (version_compare(PHP_VERSION, '8.0.0') >= 0) {
+                $dependencies[] = $container->make($param->getType()->getName());
+            }
+            else{
+                $dependencies[] = $container-make($param->getClass()->name);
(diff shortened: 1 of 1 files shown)


zhooravell:
We need this fix))

avoronin:
My project depends on this fix since we are using php8. Please let me know if I can help get this fix sooner. Thanks in advance!

Free4evo:
Keep this fix, please

nkoptel:
Please merge this fix

irazasyed:
Thanks

src/Answers/AnswerBus.php
```diff
@@ -57,7 +57,12 @@ protected function buildDependencyInjectedAnswer($answerClass)
         $container = $this->telegram->getContainer();
         $dependencies = [];
         foreach ($params as $param) {
-            $dependencies[] = $container->make($param->getClass()->name);
+            if (version_compare(PHP_VERSION, '8.0.0') >= 0) {
+                $dependencies[] = $container->make($param->getType()->getName());
+            }
+            else{
+                $dependencies[] = $container-make($param->getClass()->name);
```
andriichuk, line 64: Please check arrow sign (->)

semsphy: it is fixed in #870

Issue #837, opened 2021-01-27
Aliased command not received arguments


ddzobov:
```
class LoginCommand extends Command
{
    /**
     * @var string Command Name
     */
    protected $name = 'login';

    /**
     * @var string Command Aliases
     */
    protected $aliases = ['start'];

    /**
     * @var string Command Pattern
     */
    protected $pattern = '{code}';

    /**
     * @var string Command Description
     */
    protected $description = 'Bot login';

    public function handle(): void
    {
        $arguments = $this->getArguments();
        app('log')->info('args', $arguments);
    }
}
```

Cases:
`/login somecode` => i see "somecode" in logs
`/start somecode` => i see empty array in logs

PR #857, opened 2021-04-16
Fix: Aliased command not received arguments

arispati:
I've the same issue with #837 that using alias command not received arguments.

After checking, it turns out that the problem is in the regex in the prepareRegex function in the Telegram\Bot\Commands\Command class.

Example:
```
class ExampleCommand extends Command
{
    protected $name = 'example';
    protected $aliases = ['e'];
    protected $pattern = '{code}';
}
```

`/example text`
> `%/example(?:@.+?bot)?\s+?(?P<code>[^ ]++)%si`

`/e text`
> `%/example(?:@.+?bot)?\s+?(?P<code>[^ ]++)%si`

as you can see, the regex use command name instead of the alias name

### DIFF
--- a/src/Commands/Command.php
+++ b/src/Commands/Command.php
@@ -271,7 +271,14 @@ abstract class Command implements CommandInterface
             })
             ->implode('');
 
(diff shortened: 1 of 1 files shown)


irazasyed:
Thanks

PR #864, opened 2021-05-24
createChatInviteLink Method Add

fabiosantanarp:
Guys, I coded this method. I'm not able to send this modify now. But I post method here. Please, update lib, if convenient:

`
 /**
     * Create an additional invite link for a chat.
     *
     * The bot must be an administrator in the group for this to work.
     *
     * <code>
     * $params = [
     *   'chat_id'              => ''
     *   'expire_date'          => ''
     *   'member_limit'         => ''
     * ];
     * </code>
     *
     * @link https://core.telegram.org/bots/api#createChatInviteLink
     *
     * @param array    $params  [
     *
     * @var string|int $chat_id Unique identifier for the target chat or username of the target channel (in the format @channelusername)
     * @var int $expire_date Point in time (Unix timestamp) when the link will expire
     * @var int $member_limit Maximum number of users that can be members of the chat simultaneously after joining the chat via this invite link; 1-99999
     *
     * ]
     *
     *
     * @return string
     */
    public function createChatInviteLink(array $params): array
    {
        $response = $this->post('createChatInviteLink', $params);

        return $response->getResult();
    } `


irazasyed:
This method is implemented in v4. Thanks for the PR tho.

PR #866, opened 2021-06-02
Added ability to use custom command processors.

butschster:
There was only one way to match command with telegram message via CommandBus and it was hardcoded.

I separated command matching logic with command bus and now users can create their own processors and combine them.

Processors can be set in config file.

**For example**

Let's imagine that we have a command for storing tweets from telegram bot in database. And we want to send just tweet url without command `/handle_tweet ....`, only url.

```php
class TweetHandlerCommand extends Command {
     protected $name = 'handle_tweet';

     public function handle() 
     {
          // store tweet in database
          return true;
     }
}

```
And we can create processor, that will listen to telegram bot messages and match them with regex and route to `TweetHandlerCommand`

```php
use Telegram\Bot\Commands\CommandsProcessor;

class TweetCommandProcessor  implements CommandsProcessor {
    private const REGEX = '#https?://twitter\.com/(?:\#!/)?(?<username>\w+)/status(es)?/(?<id>\d+)#is';

    public function handle(Update $update): array
    {
        $message = $update->getMessage();

        if ($message->has('entities')) {
            return collect($message->get('entities'))
                ->filter(unction (array $entity) { return preg_match(self::REGEX, $message->text, $match) !== 1})
                ->mapWithKeys(function (array $entity) use ($parser, $update) {
                    return ['handle_tweet' => $entity];
                })->all();
        }

        return [];
    }
}
```

or

```php

use Telegram\Bot\Answers\Answerable;
use Telegram\Bot\Api;
use Telegram\Bot\Commands\CommandsProcessor;

class TweetCommandProcessor  implements CommandsProcessor {
    use Answerable;

    private const REGEX = '#https?://twitter\.com/(?:\#!/)?(?<username>\w+)/status(es)?/(?<id>\d+)#is';

    public function handle(Update $update): array
    {
        $message = $update->getMessage();

        if ($message->has('entities')) {
            return collect($message->get('entities'))
                ->filter(unction (array $entity) { return preg_match(self::REGEX, $message->text, $match) !== 1})
                ->map(function (array $entity) use ($update) {
                    return function(Api $telegram, Update $update) {
                           $this->setTelegram($telegram);
                           $this->update = $update;

                           // Store tweet in database 
                           $this->replyWithMessage([
                                'text' => 'Tweet stored in database',
                          ]);

                           return true;
                    };
                })->all();
        }

        return [];
    }
}
```

or maybe you want to log messages

```php
use Telegram\Bot\Commands\CommandsProcessor;

class LogMessagesCommandProcessor  implements CommandsProcessor {

    public function handle(Update $update): array
    {
        $message = $update->getMessage();


        return [function(Api $telegram, Update $update) use($message) {
                 logger()->debug('new tweet', [
                     'text' => $message->text
                 ]);
       }];
    }
}
```

I think it will be a good feature for your v3.x

### DIFF
--- a/src/Api.php
+++ b/src/Api.php
@@ -4,6 +4,7 @@ namespace Telegram\Bot;
 
 use BadMethodCallException;
 use Illuminate\Support\Traits\Macroable;
+use Telegram\Bot\Commands\CommandBus;
 use Telegram\Bot\Exceptions\TelegramSDKException;
 use Telegram\Bot\HttpClients\HttpClientInterface;
 
@@ -45,14 +46,18 @@ class Api
      * Instantiates a new Telegram super-class object.
      *
      *
+     * @param CommandBus               $commandBus        The Telegram Bot API Access Token.
      * @param string                   $token             The Telegram Bot API Access Token.
      * @param bool                     $async             (Optional) Indicates if the request to Telegram will be asynchronous (non-blocking).
      * @param HttpClientInterface|null $httpClientHandler (Optional) Custom HTTP Client Handler.
      *
      * @throws TelegramSDKException
      */
-    public function __construct($token = null, $async = false, $httpClientHandler = null)
+    public function __construct(CommandBus $commandBus, $token = null, $async = false, $httpClientHandler = null)
     {
+        $commandBus->setTelegram($this);
+        $this->commandBus = $commandBus;
+
         $this->accessToken = $token ?? getenv(static::BOT_TOKEN_ENV_NAME);
         $this->validateAccessToken();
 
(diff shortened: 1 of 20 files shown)


irazasyed:
While this looks like a great idea, I think this matches with our events system that is built into the upcoming v4 release already and has wide scope for various use cases. Appreciate your PR tho.

You can explore the V4 code base [here](https://github.com/telegram-bot-sdk/telegram-bot-sdk) or [Laravel](https://github.com/telegram-bot-sdk/laravel) package. Feel free to try the project and contribute with your suggestions and PRs there.

Thanks!

PR #867, opened 2021-06-02
add deleteMessage method

leonardooleg:
Add method deleteMessage  in Message.php
A very useful feature for the Telegram bot


irazasyed:
It's already [available](https://github.com/irazasyed/telegram-bot-sdk/blob/develop/src/Methods/EditMessage.php#L229-L234). Thanks.

Issue #394, opened 2017-05-18
problem with send photo


iMohammadd:
hello
i want send photo stored on my server bot getting error:

> CouldNotUploadInputFile in CouldNotUploadInputFile.php line 29: Filename not provided for http://domain.com/XXXXX.jpg. Remote or Resource file uploads require a filename. Refer Docs for more information.

my code:
`$this->telegram->sendPhoto([
            'chat_id'=>'XXXXXX',
            'photo'=>'http://domain.com/XXXXX.jpg'
 ]);`

firegore2:
Did you try using https url?

iMohammadd:
@firegore2
yes

jonnywilliamson:
Latest version (develop branch) must use an `InputFile` object when sending ANY files.

```
    \Telegram::bot('sdk')->sendPhoto([
        'chat_id' => '',
        'photo'   => InputFile::create($resourceOrFile, $filename),
        'caption' => '',
    ]);
```

Issue #206, opened 2016-07-19
I can't send photo


mmnaderi:
hello...
I make a photo in another file with php,
and I try to send photo:

                $response = $telegram->sendPhoto([
                    'chat_id' => -1001066838580,
                    'photo' => "https://gentle-earth-84565.herokuapp.com/distich-maker/randomDistich.php?m1=".$request->m1."&m2=".$request->m2."&poet=".$request->poet
                ]);

and the maked url like this:
`https://example-84565.herokuapp.com/distich-maker/randomDistich.php?m1=%D9%BE%DB%8C%D8%B1%20%D9%BE%DB%8C%D9%85%D8%A7%D9%86%D9%87%E2%80%8C%DA%A9%D8%B4%20%D9%85%D9%86%20%DA%A9%D9%87%20%D8%B1%D9%88%D8%A7%D9%86%D8%B4%20%D8%AE%D9%88%D8%B4%20%D8%A8%D8%A7%D8%AF&m2=%DA%AF%D9%81%D8%AA%20%D9%BE%D8%B1%D9%87%DB%8C%D8%B2%20%DA%A9%D9%86%20%D8%A7%D8%B2%20%D8%B5%D8%AD%D8%A8%D8%AA%20%D9%BE%DB%8C%D9%85%D8%A7%D9%86%E2%80%8C%D8%B4%DA%A9%D9%86%D8%A7%D9%86&poet=%D8%AD%D8%A7%D9%81%D8%B8`
the parameters is in persian language.

but this says:

> Bad Request: Wrong persistent file_id specified: contains wrong characters or has wrong length

vladimir-yuldashev:
Hello, you should be doing something like this:

`
$this->telegram()->sendPhoto([
    'photo' => new InputFile($pathToPhoto)
]);
`

mmnaderi:
I do this...
but this error:

```
ErrorException in Api.php line 983:
is_file() expects parameter 1 to be a valid path, object given
```

vladimir-yuldashev:
https://telegram-bot-sdk.readme.io/docs/sendphoto

```
use Telegram\Bot\Api;

$telegram = new Api('BOT TOKEN');

$response = $telegram->sendPhoto([
  'chat_id' => 'CHAT_ID', 
  'photo' => 'path/to/photo.jpg', 
  'caption' => 'Some caption'
]);

$messageId = $response->getMessageId();
```

Worked for me.

[...]

Issue #42, opened 2015-09-25
[request] SendPhoto from URL


aleritty:
Actually your API support only the official method (upload a local file OR reference a file_id).

But automatically handling URL (so submitting as multipart data) can be really useful.

Thank you for your work!!

irazasyed:
Well, We could add support for that but you could simply use built-in PHP functions like `file_get_contents` and pass on the image URL, then the response to the sendPhoto method.

But thanks for the suggestion though. I'll consider this before the next release and see if we can implement it.

jonnywilliamson:
+1

azizbek8434:
hi, i couldn't sent photo using this api methods
```
$response = Telegram::sendPhoto([
  'chat_id' => 'CHAT_ID', 
  'photo' => 'path/to/photo.jpg', 
	'caption' => 'Some caption'
]);
```

>and this is json data and image path

![default](https://user-images.githubusercontent.com/36416819/49513723-bf78ce80-f8b3-11e8-9030-373c1f3bffb9.png)

> and this is my code
```
    protected function getProducts($chat_id,$category_slug,$manager_slug,$lang='en'){
    $products = file_get_contents('https://test.com/api/products?manager='.$manager_slug.'&category='.$category_slug);
        $data = json_decode($products,true);
        foreach($data['data'] as $product){
        Telegram::sendPhoto([
            'chat_id' => $chat_id,
            'photo' => 'https://test.com/public'.$product['image'],
            'caption' => $product['name']
        ]);
    }
```
please help me, thanks in advance..

PR #868, opened 2021-06-08
Support upload file via URL

Sea-n:
Upload files via URL can benefit from CDN and Telegram cache and most methods support it. It could be much faster when sending the same file more than once.

Hope this library can support it too!

Ref: [Telegram Bot API Manual > Sending Files](https://core.telegram.org/bots/api#sending-files), issue #42, issue #206, issue #394.


irazasyed:
This feature is supported in upcoming v4. Thanks for the PR tho.

PR #879, opened 2021-07-19
Handle all possible Guzzle request exceptions

b1rdex:
`RequestException` is just a subset of possible exceptions thrown while performing the request.
For example, connection failure leads to `ConnectException` and so on.

The best way is to catch all `GuzzleException` instances and wrap them in `TelegramSDKException`.

Also, I added the original exception as an argument to `TelegramSDKException` so it can be retrieved in the app later.


irazasyed:
Thanks!

PR #885, opened 2021-08-16
Fixed conflict command aliases

korkoshko:
This problem occurs when we have multiple instances of the bot manager.
The bot manager initializes the bot each time and registers its commands (command bus - singleton).


irazasyed:
Thanks

PR #893, opened 2021-09-17
add IDE directories in gitignore

amir1376:
prevent add IDE (PHP Storm and VS Code ) configurations accidently into repository


irazasyed:
This should go in your global gitignore file.

PR #897, opened 2021-10-09
Update guzzlehttp/psr7 to version 2

fffaraz:
because Larevel 8 uses guzzlehttp/psr7 version 2.1.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
   "require": {
     "php": "^7.2 || ^8.0",
     "guzzlehttp/guzzle": "^6.3 || ^7.0.1",
-    "guzzlehttp/psr7": "^1.3",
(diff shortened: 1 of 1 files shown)


alex-kudrya:
plz merge it!

irazasyed:
This isn't compatible with our code for now as you can see the tests fail. Either fix them or wait for v4 release. Thanks.

fffaraz:
Seems like `\GuzzleHttp\Psr7\stream_for` should be replaced with `\GuzzleHttp\Psr7\Utils::streamFor`.
I'm not an expert in PHP but I guess I can try if it's just this.

irazasyed:
Thanks for the PR. This is fixed in #907 PR which was merged now.

PR #907, opened 2021-11-13
Psr7 v2 support

DoctorLines:
1. Fix `composer.json` require config: guzzle 6.3.0 is not support psr7 1.3.0 (need up to 1.4)
2. Add unit tests to compatible with psr7 2 and leave backward compatibility with 1.4.0
3. Add Psr7 v2 support

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
   "require": {
     "php": "^7.2 || ^8.0",
     "guzzlehttp/guzzle": "^6.3 || ^7.0.1",
-    "guzzlehttp/psr7": "^1.4",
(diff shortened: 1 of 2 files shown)


DoctorLines:
Failing because extensions cache environment collision error
```
Error: Unable to reserve cache with key Linux-focal-ext-7.4-33518cee0ed0705d142b8070ad4da489230efbcf5a49d7ead12fca99004fdc74-20210919
-deps, another job may be creating this cache.
```
Retry 4 times and success))
https://github.com/DoctorLines/telegram-bot-sdk/runs/4197408592

irazasyed:
Thanks

PR #927, opened 2022-02-26
add copyMessage

mrjavadseydi:
adding copyMessage method.
Use this method to copy messages of any kind. Service messages and invoice messages can't be copied.
[https://core.telegram.org/bots/api#copymessage](https://core.telegram.org/bots/api#copymessage)


irazasyed:
Thanks 👍🏻

PR #928, opened 2022-02-26
Added PHPDoc templates for Keyboard classes

Kravets1996:
Added PHPDoc templates for Keyboard classes

More info in this issue:
#929


irazasyed:
Thanks 👍🏻

Issue #933, opened 2022-03-20
"Telegram Bot PHP SDK v3.0.0 - (https://github.com/irazasyed/telegram-bot-sdk)" is not valid header value


andrey-helldar:
- Telegram Bot: **v3.5**
- Laravel Framework: **9.5.1**
- `guzzlehttp/http`: **7.4.1**
- `guzzlehttp/psr7`: **2.2.0** (2.1.0 - no errors)
- PHP: **8.1.4**

```json
{
    "error": {
        "type": "InvalidArgumentException",
        "data": "\"Telegram Bot PHP SDK v3.0.0 - (https:\/\/github.com\/irazasyed\/telegram-bot-sdk)\" is not valid header value"
    }
}
```

<details>
  <summary>Stack Trace</summary>
  
  ```text
[2022-03-20 14:20:43] local.ERROR: "Telegram Bot PHP SDK v3.0.0 - (https://github.com/irazasyed/telegram-bot-sdk)" is not valid header value {"exception":"[object] (InvalidArgumentException(code: 0): \"Telegram Bot PHP SDK v3.0.0 - (https://github.com/irazasyed/telegram-bot-sdk)\" is not valid header value at D:\\domains\\dragon-code\\apps\\webhooks\\vendor\\guzzlehttp\\psr7\\src\\MessageTrait.php:263)
[stacktrace]
#0 D:\\domains\\dragon-code\\apps\\webhooks\\vendor\\guzzlehttp\\psr7\\src\\MessageTrait.php(209): GuzzleHttp\\Psr7\\Request->assertValue('Telegram Bot PH...')
#1 [internal function]: GuzzleHttp\\Psr7\\Request->GuzzleHttp\\Psr7\\{closure}('Telegram Bot PH...')
#2 D:\\domains\\dragon-code\\apps\\webhooks\\vendor\\guzzlehttp\\psr7\\src\\MessageTrait.php(212): array_map(Object(Closure), Array)
#3 D:\\domains\\dragon-code\\apps\\webhooks\\vendor\\guzzlehttp\\psr7\\src\\MessageTrait.php(174): GuzzleHttp\\Psr7\\Request->trimAndValidateHeaderValues(Array)
#4 D:\\domains\\dragon-code\\apps\\webhooks\\vendor\\guzzlehttp\\psr7\\src\\MessageTrait.php(154): GuzzleHttp\\Psr7\\Request->normalizeHeaderValue('Telegram Bot PH...')
[...]

PR #934, opened 2022-03-20
Fix `is not valid header value`

andrey-helldar:
Fix for https://github.com/irazasyed/telegram-bot-sdk/issues/933


irazasyed:
Thanks!

PR #961, opened 2022-07-02
Add missing types

alies-dev:
Add missing types:

 * ChatInviteLink
 * ChatJoinRequest
 * ChatMemberUpdated
 * PollAnswer

It changes `Update::detectType()` output, that means this is BC changes, but there is nothing to do with it due to bad package architecture.


irazasyed:
Thanks!

PR #964, opened 2022-07-04
Fix one-to-many relations

alies-dev:
## Problem

We have one-to-many relations, e.g. Message has few Photos. For both one-to-one and one-to-many relations, the syntax is the same:
```php
public function relations()
{
    return [
        'photo' => PhotoSize::class,
    ];
}
```

the code to return value of both types of relation is here:
```php
if (isset($relations[$property])) {
    return $relations[$property]::make($value);
}
```
Where:
 - `$relations[$property]` -- FQCN of the BaseObject child that extends Collection
 - `$value` -- relation's raw data (json parsed into associative array)

It works perfectly in case of one-to-one (e.g. Message may have only a single Document, so it will call `Document::make($value)` that will return `Document` instance).

In case of one-to-many relation type, it will call PhotoSize::make($value), but that $value - is data for ARRAY of PhotoSize objects. As the result, $message->photo will return PhotoSize instance instead of a list of PhotoSize instances.

## Solution
Use array syntax to specify one-to-many relation:
```diff
public function relations()
{
    return [
-        'photo' => PhotoSize::class,
+        'photo' => [PhotoSize::class],
    ];
}
```

Plus update a method to initiate relation from raw data (see a new `getRelationValue` method).

## Notes

IMO this is not a breaking compatibility change, but changes big enough for a new minor version.


irazasyed:
This is interesting. Did you try our upcoming [V4](https://github.com/telegram-bot-sdk/telegram-bot-sdk) and see if this problem exists there too?

alies-dev:
@irazasyed 
Nope, I'm currently using v3 for my project. There was no info about v4 on this repo and I was sure that 3.6 is the latest one :/

irazasyed:
@lptn That's because it's been in WIP status for quite some time and it's a major change yet to be released. Many things decoupled and events-driven architecture. You should explore and see :)

jonnywilliamson:
@lptn You should pop over to V4 of this library and if have a look. https://github.com/telegram-bot-sdk/telegram-bot-sdk

We could always use some new eyes and help to get it fully released!

alies-dev:
@irazasyed 
@jonnywilliamson 

Thanks guys, I really appreciate your work and contributions! I see a problem, let me share it (I'm trying to be constructive as possible, underline problems, but not blame anyone -- we are open-source voluntaries only).

In a nutshell this problem -- it work with community. There are docs, but they are for outdated v2. There are v3, but there are no docs and no published releases on GitHub (only pre-releases). You migrated to another repository, but current repository haven't linked into a new one. You suggest me to use v4, but it explicitly says `[WIP - DO NOT USE IN PRODUCTION YET]` (while I'm building production-ready bot). I couldn't find docs on how to migrate from v3 to v4 (v4 docs leads to a broken domain: https://telegrambotsdk.com/). Latest commit on v4 repo dated by Jan 28 (almost half of year ago). There are no new v4 git tags or releases on v4 repo (even alpha/beta/RC/pre-release ones).

I'm sure there are reasons of having these issues, but at least some of the are pretty easy to fix.

All these listed factors make v4 release/repo not really reliable (at lease for my fresh eyes). I like solutions and changes you applied for v4, but without proper docs/marketing/positioning , v4 will be pretty unfriendly for new comers like me.


Few days ago I was not aware about v4 and even built a Conversations plugin for v3: https://github.com/koot-labs/telegram-bot-dialogs. Now, in order to move to v4, I need to update both my project and this plugin to work, what will require some time that I can spend improving my bots(s) instead.

PR #966, opened 2022-07-07
Drop PHP 7.2 support and update phpunit

alies-dev:
Goals:
 - fix tests on CI (phpunit 8.X can't be installed on php 7.2)
 - help PHP ecosystem by forcing developers using supported PHP version (all versions older 7.4 are [EOL](https://www.php.net/supported-versions.php))

PR #967, opened 2022-07-07
Fix type issue parsing "entities"

alies-dev:
## Before review
It's better to firstly merge https://github.com/irazasyed/telegram-bot-sdk/pull/966 (where I've fixed tests on CI). Tests failing for a long time, it's not caused by this PR

## Description
Because we fixed one-to-many relationships, now we have collections on objects instead of collections of arrays.

Currently I've located a related bug caused by a change in this line: https://github.com/koot-labs/telegram-bot-sdk-v3/commit/55c2b7e2b8b5f850a646948e042bb79b8e46a8ae#diff-57d0f563404fb108cfd0cbe4edfec9b74c6ad896942dcff31be394e8dd7461e4R74

the fix is ready and tested on a live bot

### DIFF
--- a/src/Commands/Command.php
+++ b/src/Commands/Command.php
@@ -5,6 +5,7 @@ namespace Telegram\Bot\Commands;
 use Illuminate\Support\Collection;
 use Telegram\Bot\Answers\Answerable;
 use Telegram\Bot\Api;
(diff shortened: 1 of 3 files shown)


irazasyed:
Thanks 👍

PR #974, opened 2022-07-08
Add backward compatibility check to CI (backported from SDK v4)

alies-dev:
I think it's beneficial to use it on this repository also

### DIFF
--- a/.github/workflows/backward-compatibility-check.yml
+++ b/.github/workflows/backward-compatibility-check.yml
@@ -1,0 +1,15 @@
+on: [push]
+name: Backward compatibility check
+
+jobs:
+  bc-check:
+    name: Backward compatibility check
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


alies-dev:
**UPD:** on my fork, this GitHub workflow wrote `Detected last minor version: v3.4.0`. Because of this, it generates a lot of noise: v3.4 used Laravel 8, that used Collections without return and param types and TG Objects in this SDK extends Collection class. I hope on this repository it will correctly detect latest minor version to compare HEAD with. 🤞 If not -- we can easily remove this workflow (on other repositories where I use it ([example](https://github.com/spatie/calendar-links/actions/workflows/backward-compatibility-check.yml)) I don't have such issues)

irazasyed:
I think we can skip this one for this repo since we don't plan to support V3 for long nor this repo will be active once V4 migration is done.

PR #985, opened 2022-07-19
Fix return type error on method findType

MaizerGomes:
This pull request fixes the following error on php7.4:
 `Return value of Telegram\Bot\Objects\BaseObject::findType() must be of the type string or null, none returned`

This error would not allow the method get-updates to run.


irazasyed:
Thanks

PR #996, opened 2022-08-09
Add new `ApiRequestSending` and `ApiResponseReceived` events

alies-dev:
This adds 2 new events:
 - `ApiRequestSending`
 - `ApiResponseReceived`

For far the package provides only a single event: `UpdateWasReceived`. Having the only one event it's possible to use event system for logging and other purposes - you don't have the full picture. These 2 new events is simple but powerful addition. 

I'm open to:
 - rename them
 - port them to SDK v4 (my long-term goal is to simplify migration to SDK v4)
 - remove `if (method_exists($this, 'emitEvent')) {` part -- I found it ugly, but there is no way to make code safe without interfaces when you use traits in that fashion.

### DIFF
--- a/src/Events/ApiRequestSending.php
+++ b/src/Events/ApiRequestSending.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Telegram\Bot\Events;
+
+use League\Event\AbstractEvent;
+use Telegram\Bot\TelegramRequest;
+
+final class ApiRequestSending extends AbstractEvent
+{
+    /** @var \Telegram\Bot\TelegramRequest */
+    private $request;
+
+    public function __construct(TelegramRequest $request)
+    {
+        $this->request = $request;
+    }
+
+    public function getRequest(): TelegramRequest
+    {
+        return $this->request;
+    }
+}
--- a/src/Events/ApiResponseReceived.php
+++ b/src/Events/ApiResponseReceived.php
@@ -1,0 +1,22 @@
+<?php declare(strict_types=1);
+
+namespace Telegram\Bot\Events;
+
+use League\Event\AbstractEvent;
+use Telegram\Bot\TelegramResponse;
+
+final class ApiResponseReceived extends AbstractEvent
+{
+    /** @var \Telegram\Bot\TelegramResponse */
+    private $response;
+
+    public function __construct(TelegramResponse $response)
+    {
+        $this->response = $response;
+    }
+
+    public function getResponse(): TelegramResponse
+    {
+        return $this->response;
+    }
+}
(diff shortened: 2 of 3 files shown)


irazasyed:
Have you explored V4? It's all event-driven and pretty much for every inward messages, it fires various events (from update to the type of message, etc.).

While this addition here seems simpler but may not really be that useful at this time considering this version isn't heavy on events plus v4 uses a different events system.

I feel we should focus on improving V4 and then the pending docs, to make things easier to transit. Thoughts?

alies-dev:
>I feel we should focus on improving V4 and then the pending docs, to make things easier to transit. Thoughts?

## events
Let me explore events in v4, probably I can port it to v3. Also, I already have written and use an adapter from League events to Laravel events on my project, so this is not a problem to me (I can open-source it for the community on demand).


## migration to v4
I don't like some architecture decisions on v3 and especially cameCased properties/attributes, but I'm too lazy to migrate my not small project to v4: v3 works for me well, I've fixed all issues I faced, I even wrote my own router (that I can also share) for this version to replace command bus completely.
SDK v4 looks not friendly without at least migration docs. Also, there is no clean way how it's possible to contribute to docs for version v4 (however the same for v3 :) ), but at least docs for SDK v2 partially works for v3, this is why a lot of devs chose v3. Another reason - there is no clean message that this version is archived and you are working on v4. Taking into account work with community for SDKv4, a complete lack of docs and low contribution activity (as result) -- I think it's safer to stay on SDKv3 as it looks more reliable and bottle-tested, at least for the outside visitor like me.
 Required minimum to me to start migration:
 - have basic migration docs 
 - have a simple way to contribute to migration docs

alies-dev:
@irazasyed 
I'll try to backport events from v4, we can close this PR. But I'll wait for your feedback for my previous message.

irazasyed:
V4 lacks contribution because we neither officially linked from this repo nor really promoted it here ever due to lack of docs. Initially, we thought to finish the docs and then finally go live but unfortunately, everyone in the contributor's team got caught up with various things in our lives and that didn't go through as planned. It has been under work among a set of contributors from the community and it's a major shift with some really crazy features that I'm sure you'd love once we go live as it makes the bot development experience super easy with magic params and stuff.

I understand migration docs are needed, I'll look into this as soon as I get some time. I'll invite you to the team over there so you can explore further and see if you find anything that needs improvements. 

I'll also look into setting up a docs site so we can work on it later.

Thanks for all your PRs and improvements. I appreciate it.

V3 will be around for quite some time until V4 is fully released and mature enough to replace.

PR #998, opened 2022-08-09
Emit `UpdateEvent` event similar to SDK v4

alies-dev:
`UpdateEvent` event back-ported from SDK v4.

Differences with SDKv4 events:
 - event from SDKv4 has `public Bot $bot`, in current PR it's `public Api $telegram;`. Reason -- there is no `Bot` class in SDK v3.
 - in this PR this event extends `\League\Event\AbstractEvent` (because we use `league/event` events system here), in SDK v4 it doesn't extend any classes.
 - this PR has better test than in SDK v4 :)

## Challenges

SDK v3 has a weird feature, I don't know a reason why we have it. This is `$shouldEmitEvent` parameter:
```php
// in Update.php trait
public function getWebhookUpdate($shouldEmitEvent = true, ...
```

I decided to not touch it, always emit a new event and keep emitting the old one `UpdateWasReceived` depends on $shouldEmitEvent argument's value. Reason - it provides more consistency with SDK v4.

### DIFF
--- a/src/Methods/Update.php
+++ b/src/Methods/Update.php
@@ -236,13 +236,13 @@ trait Update
 
         $event = new UpdateEvent($this, $update);
 
-        $eventEmitter->emit($event->cloneWithCustomName(UpdateEvent::NAME));
+        $eventEmitter->emit(UpdateEvent::NAME, $event);
         $updateType = $update->objectType();
         if (is_string($updateType)) {
-            $eventEmitter->emit($event->cloneWithCustomName($updateType));
+            $eventEmitter->emit($update->objectType(), $event);
 
             if (null !== $update->getMessage()->objectType()) {
-                $eventEmitter->emit($event->cloneWithCustomName($updateType . '.' . $update->getMessage()->objectType()));
+                $eventEmitter->emit($update->objectType() . '.' . $update->getMessage()->objectType(), $event);
             }
         }
     }
--- a/src/Methods/Update.php
+++ b/src/Methods/Update.php
@@ -234,15 +234,14 @@ trait Update
 
         $eventEmitter = $this->eventEmitter;
 
-        $event = new UpdateEvent($this, $update);
-
-        $eventEmitter->emit(UpdateEvent::NAME, $event);
+        $eventEmitter->emit(new UpdateEvent($this, $update));
         $updateType = $update->objectType();
         if (is_string($updateType)) {
-            $eventEmitter->emit($update->objectType(), $event);
+            $eventEmitter->emit(new UpdateEvent($this, $update, $updateType));
 
-            if (null !== $update->getMessage()->objectType()) {
-                $eventEmitter->emit($update->objectType() . '.' . $update->getMessage()->objectType(), $event);
+            $messageType = $update->getMessage()->objectType();
+            if (null !== $messageType) {
+                $eventEmitter->emit(new UpdateEvent($this, $update, "$updateType.$messageType"));
             }
         }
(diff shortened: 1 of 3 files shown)


src/Methods/Update.php
```diff
@@ -220,4 +224,26 @@ private function getRequestBody(?RequestInterface $request)
...
+
+        $eventEmitter->emit($event->cloneWithCustomName(UpdateEvent::NAME));
+        $updateType = $update->objectType();
+        if (is_string($updateType)) {
+            $eventEmitter->emit($event->cloneWithCustomName($updateType));
+
+            if (null !== $update->getMessage()->objectType()) {
+                $eventEmitter->emit($event->cloneWithCustomName($updateType . '.' . $update->getMessage()->objectType()));
+            }
+        }
```
lptn, line 247: side by side comparison with SDK v4 (if it will help you to review):
image

src/Methods/Update.php
```diff
@@ -220,4 +224,26 @@ private function getRequestBody(?RequestInterface $request)
...
+
+        $eventEmitter = $this->eventEmitter;
+
+        $event = new UpdateEvent($this, $update);
+
+        $eventEmitter->emit($event->cloneWithCustomName(UpdateEvent::NAME));
+        $updateType = $update->objectType();
+        if (is_string($updateType)) {
+            $eventEmitter->emit($event->cloneWithCustomName($updateType));
```
irazasyed: Why is the UpdateEvent being reused here? In V4, we dispatch 2-3 different events. Update, Update Type/Sub Message event. We don't use a class, it's simply a string.

alies-dev: Hm, makes sense, honestly I thought that league/events v2 supports only class-events without name. I'll update the PR tomorrow 👍

irazasyed: Thanks!

alies-dev: I just checked docs and I was right: when you emit an event using `league/events v2` it's not possible to specify name and event as separate arguments, only like this:
```php
$emitter->emit(new UpdateEvent());
// or
$emitter->emit('string.event');
```

But it's possible to provide extra arguments: https://event.thephpleague.com/2.0/events/arguments/


I'll try this option. It's sort of a hack, but probably will work for this case.

alies-dev: @irazasyed 
your goal is to use the same even names as on SDK v4, right? If so, it already worked (and works), but I've added/improved a test to make sure. We don't even need to use "arguments" feature from league/events v2

tests/Integration/TelegramApiTest.php
```diff
@@ -543,7 +552,34 @@ public function check_the_webhook_works_and_can_emmit_an_event()
...
+
+        $api->getWebhookUpdate(true, $incomeWebhookRequest);
+        $allEvents = $listener->events;
+
+        $this->assertArrayHasKey(UpdateEvent::NAME, $allEvents);
+        $this->assertArrayHasKey('message', $allEvents);
+        $this->assertArrayHasKey('message.text', $allEvents);
+        $this->assertCount(1, $allEvents[UpdateEvent::NAME]);
+        $this->assertCount(1, $allEvents['message']);
+        $this->assertCount(1, $allEvents['message.text']);
```
alies-dev, line 582: 3 events with same names as on SDK v4, event class has the same public methods (except it has `Api` instead `Bot` that is missing in SDK v3)

irazasyed:
Thank you 👍

PR #999, opened 2022-08-12
fix chat trait

kolirt:


.gitignore
```diff
@@ -1,3 +1,4 @@
+/.idea
```
irazasyed: This should go into your global ignores list. Please remove it from here.

kolirt: removed

irazasyed:
Thanks!

PR #1011, opened 2022-09-29
Fix false positives for `/help` command

andrey-helldar:
For example, the path to the `/storage` folder is sent to the chat. Telegram defines `/storage` as a command, and the SDK, when it doesn't find one, calls the `/help` command because it was added to the bot.

It turns out that the bot reacts to someone else's command.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,7 @@
   },
   "require-dev": {
     "php-parallel-lint/php-parallel-lint": "^1.3",
+    "phpspec/prophecy": "^1.15",
     "phpunit/phpunit": "^9.5"
   },
   "autoload": {
--- a/src/Commands/CommandBus.php
+++ b/src/Commands/CommandBus.php
@@ -224,7 +224,6 @@ class CommandBus extends AnswerBus
     {
         $command = $this->commands[$name] ??
             $this->commandAliases[$name] ??
-            $this->commands['help'] ??
             collect($this->commands)->filter(function ($command) use ($name) {
                 return $command instanceof $name;
             })->first() ?? null;


irazasyed:
That's the default behavior of the SDK. Many of the bots I know rely on this as the fallback command in cases of all commands failing.

I'm concerned this may break many bots. The newer version of the SDK already lets you have full control to enable/disable this command from the config.

andrey-helldar:
I agree that these are major changes. However, using version `3.9` I ran into this problem.

Part of my config file:

```php
use App\Enums\Telegram\BotName;

return [
    'bots' => [

        BotName::CLEANER() => [
            'username' => BotName::CLEANER(),

            'token' => env('TELEGRAM_LIGHT_CLEANER_BOT_TOKEN'),

            'commands' => [
                Start::class,
                Report::class,
                Ban::class,
                MessageId::class,
                Validate::class,
                Help::class,
            ],
        ],
    ],

    'commands' => [],
    'command_groups' => [],
    'shared_commands' => [],
];
```

If you send, for example, `/storage` to a chat with a bot, then the `/help` command will be called.

This is also non-obvious and incorrect behavior, because `/storage` is a folder in a Laravel project...

```json
{
  "message": {
    "chat": {
      "id": -1001712003365,
      "type": "supergroup",
      "title": "The Dragon Code Test"
    },
    "date": 1664480773,
    "from": {
      "id": 123,
      "is_bot": false,
      "last_name": "Helldar",
      "first_name": "Andrey",
      "language_code": "ru"
    },
    "text": "For example, /storage",
    "entities": [
      {
        "type": "bot_command",
        "length": 8,
        "offset": 13
      }
    ],
    "message_id": 1371
  },
  "update_id": 616020739
}
```
// run the `help` command

andrey-helldar:
By the way, I just noticed one more problem: if you remove the declaration of the `/help` command in the configuration file and send any command that is not declared to the chat, the SDK will transfer control to the first one found in the list.

https://github.com/irazasyed/telegram-bot-sdk/blob/022689b92dfe05406b895f1a19f1404ba2ef5fd1/src/Commands/CommandBus.php#L223-L233

andrey-helldar:
I saw a new project with [version 4](https://github.com/telegram-bot-sdk/telegram-bot-sdk).

I close the current PR, because it makes no sense.

PR #1029, opened 2023-01-19
Fix WebhookCommand --info parameter.

sowich:
When running the command:
`php artisan telegram:webhook --info`

we can see the error: `Telegram\Bot\Laravel\Artisan\WebhookCommand::makeWebhookInfoResponse(): Argument #2 ($bot) must be of type string, null given, called in...`


irazasyed:
This has been improved in `3.x` branch.

PR #1038, opened 2023-03-02
Use a better issue template (with required fields)

alies-dev:
Goal: mark some mandatory fields as required - PHP and package version.

It will look like:
image

This should require a number of non-informative bug-reports.

_____

PS: it seems like we need to tag a new release, as ppl still have some issues with Laravel 10 and other dependencies - they use the latest release


irazasyed:
This is good. Thanks. 

Will tag a release shortly.

PR #1054, opened 2023-04-03
Update TelegramRequest.php

irpcpro:
in construct when properties could be null, type of these setter methods value should be also include type of null. in other case these errors will throw :
Argument #1 ($accessToken) must be of type string, null given
Argument #1 ($endpoint) must be of type string, null given
Argument #1 ($method) must be of type string, null given

### DIFF
--- a/src/TelegramRequest.php
+++ b/src/TelegramRequest.php
@@ -15,10 +15,10 @@ final class TelegramRequest
     private ?string $accessToken;
 
     /** @var string The HTTP method for this request. */
-    private string $method;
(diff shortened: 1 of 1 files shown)


irazasyed:
Will this work in the required minimum PHP version?

irpcpro:
in this file, there was type definition on each method inputs variable, and the type of property accessToken is already '?string'.
I've just added another type in each definition in this version & branch.

irazasyed:
Yes. I merged it after checking. Thanks :)

PR #1056, opened 2023-04-08
command step by step

irpcpro:
command step by step. prevent user to execute commands those are out of step map. after reject the command, class will execute failStepAction method which developer should handle this step.

### DIFF
--- a/src/Commands/CommandStepByStep.php
+++ b/src/Commands/CommandStepByStep.php
@@ -1,0 +1,144 @@
+<?php
+
+namespace Telegram\Bot\Commands;
+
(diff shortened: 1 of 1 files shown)


irazasyed:
Looks like you're trying to build a conversation system out of commands. While this seems like a good start, I encourage you to explore [v4](https://github.com/telegram-bot-sdk/telegram-bot-sdk) which is WIP but supports plugins and events. 

We do have conversation support on our roadmap but you can try and see if you can contribute a plugin for it. For now, we're not accepting any major changes to v3. But thanks for the PR, tho.

PR #1057, opened 2023-04-08
Update TelegramResponse.php

irpcpro:


irazasyed:
Can you share more details? Where is this issue happening?

irpcpro:
i've set proxy on guzzle client, but i don't know why my body response is empty.
i get http response 200 but body is empty.
i don't know why !?
i will call https://api.telegram.org/xxx/getMe separately from my browser, i get my bot details.
i don't know what is my problem exactly, i'm wokring on it.
but in this issue, when body is empty, this line will throw an error for get index result.

PR #1060, opened 2023-04-14
Update Command.php

irpcpro:
when i'm sending an image from bot, and i'm gonna trigger a command by using this :
`Telegram::triggerCommand()`
this error will throw :
Telegram\Bot\Commands\Command::relevantMessageSubString(): Return value must be of type string|bool, null returned

it because of the method type of relevantMessageSubString
by adding the type of null to this method, it will fix.

### DIFF
--- a/src/Commands/Command.php
+++ b/src/Commands/Command.php
@@ -195,7 +195,7 @@ abstract class Command implements CommandInterface
         return [$pattern, $patterns->keys()->all()];
     }
 
-    private function relevantMessageSubString(): bool|string
+    private function relevantMessageSubString(): bool|string|null
     {
         //Get all the bot_command offsets in the Update object
         $commandOffsets = $this->allCommandOffsets();


irazasyed:
You are sending an image from your command or generally?

And why is the trigger command being called outside the command system? That was not the purpose of that method.

irpcpro:
i'm sending image from telegram to bot.
and on server side, i cached previous command, and whenever user send a photo, i'm gonna trigger the next command which should be executed.

for an example, user wants to upload avatar after hitting `/avatar_upload`
the next command which should be, it's `/avatar_upload_set`
after hitting `/avatar_upload` by user, i'm gonna cache the next command `/avatar_upload_set`
whenever a photo type message will received, i'm gonna trigger the cached command which is `/avatar_upload_set`

irazasyed:
Try the updated changes. It should resolve your problem. If it still persists, please create an issue ticket.

PR #1063, opened 2023-04-15
Add a setup method for the "is_persistent" field of the ReplyKeyboardMarkup object

pavel-mironchik:
This PR just "adds" a convenient method for setting the [is_persistent](https://core.telegram.org/bots/api#replykeyboardmarkup) field which was [introduced](https://core.telegram.org/bots/api#december-30-2022) in the Bot API 6.4:

> Added the field is_persistent to the class [ReplyKeyboardMarkup](https://core.telegram.org/bots/api#replykeyboardmarkup), allowing to control when the keyboard is shown.

**P.S.** One could already use the `setIsPersistent()` method, but with this change, it will be shown by the autocomplete system of an IDE.

### DIFF
--- a/src/Keyboard/Keyboard.php
+++ b/src/Keyboard/Keyboard.php
@@ -11,6 +11,7 @@ use Telegram\Bot\Objects\LoginUrl;
  * // For Standard Keyboard
  * $params = [
  *   'keyboard'          => '',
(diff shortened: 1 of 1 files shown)


irazasyed:
Thanks

PR #1100, opened 2023-11-05
Allow to add commands stored in an iterable

romm:
Adding several commands with a single method call can be really useful, especially when working with a framework that can detect all commands instances and inject them in the command bus automatically.

This commit allows commands registration using an iterable, enhancing the possibilities to — for instance — lazy-load the commands only when the SDK API is actually called.

Below is an example of how the `Api` class is declared in a recent Symfony application:

```php
return static function (ContainerConfigurator $container): void {
    $container
        ->services()

        // Tagging all commands, so they can be retrieved below
        ->instanceof(\Telegram\Bot\Commands\CommandInterface::class)
        ->tag('telegram.command')

        // Registering the `Api` class
        ->set(Api::class)
        ->args([
            '$token' => env('TELEGRAM_BOT_TOKEN')->resolve(),
        ])
        // Adding all commands that were previously tagged
        ->call('addCommands', [tagged_iterator('telegram.command')]);
};
```

The example above would not work without changing the type of the parameter, because the `tagged_iterator` is not an array.

### DIFF
--- a/src/Commands/CommandBus.php
+++ b/src/Commands/CommandBus.php
@@ -49,11 +49,11 @@ class CommandBus extends AnswerBus
     /**
      * Add a list of commands.
      *
-     * @param  list<CommandInterface|class-string<CommandInterface>>  $commands
+     * @param iterable<CommandInterface|class-string<CommandInterface>>  $commands
      *
      * @throws TelegramSDKException
      */
-    public function addCommands(array $commands): self
+    public function addCommands(iterable $commands): self
     {
         foreach ($commands as $command) {
             $this->addCommand($command);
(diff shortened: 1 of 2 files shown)


irazasyed:
Interesting change, thanks for the PR.

romm:
Hey there, thanks for the merge!

I've been waiting for a new release and I noticed the previous one is quite old. Would it be ok to publish a new one? No urge btw, I know that OSS can be very time-consuming. 😉 

Thanks for your work on the library!

irazasyed:
@romm Thank you for your patience. Before tagging a release, I need to test it with other changes to avoid potential issues for projects that rely on this SDK. Your understanding and patience during this time is appreciated.

romm:
Understood! Wishing you the best on the upcoming tests then. 🤞

irazasyed:
https://github.com/irazasyed/telegram-bot-sdk/releases/tag/v3.14.0

PR #1115, opened 2024-03-02
allowed_updates webhook

AgelxNash:
This update allows you to control updates received from web-hook

https://core.telegram.org/bots/api#setwebhook


irazasyed:
Thanks for the PR. Please add missing tests.

PR #1135, opened 2024-08-09
Enhance Update Class to Support Command Extraction from "callback_query"

miladkhodaverdi23:
This pull request introduces a new feature to the Update class in the Telegram Bot SDK, allowing for the extraction of commands from `callback_query` updates. The enhancement adds a `getText` method that processes and retrieves command strings from `callback queries`, improving command handling efficiency.

Changes
New `getText` Method: Added to the Update class to handle command extraction from `callback_query` updates.

Usage:
Example Keyboard
You can use inline keyboards to send callback queries:
```php
Keyboard::make()
    ->inline()
    ->row([
        Keyboard::inlineButton([
            'text' => 'Access',
            'callback_data'  => 'access',
        ])
    ]);
```

Example Command Usage
Command class utilizing the new feature:

```php
class CallbackAccessAccountCommand extends Command implements CommandInterface
{
    protected string $name = 'callback:::access';
    protected string $description = 'Access account';

    /**
     * @throws Exception
     */
    public function handle()
    {
        // Command implementation
    }
}
```

Please review the changes and provide any feedback or suggestions.

Thank you!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "irazasyed/telegram-bot-sdk",
+    "name": "miladkhodaverdi23/telegram-bot-sdk",
     "description": "The Unofficial Telegram Bot API PHP SDK",
     "license": "BSD-3-Clause",
     "type": "library",
--- a/src/Commands/Command.php
+++ b/src/Commands/Command.php
@@ -165,6 +165,7 @@ abstract class Command implements CommandInterface
             return [];
         }
 
+
         // Generate the regex needed to search for this pattern
         [$pattern, $arguments] = $this->makeRegexPattern();
 
@@ -206,7 +207,7 @@ abstract class Command implements CommandInterface
         $commandOffsets = $this->allCommandOffsets();
 
         if ($commandOffsets->isEmpty()) {
-            return $this->getUpdate()->getMessage()->text ?? '';
+            return $this->getUpdate()->getText() ?? '';
         }
 
         //Extract the current offset for this command and, if it exists, the offset of the NEXT bot_command entity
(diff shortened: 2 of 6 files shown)


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "irazasyed/telegram-bot-sdk",
+    "name": "miladkhodaverdi23/telegram-bot-sdk",
```
irazasyed, line 2: Revert this change

src/Commands/Command.php
```diff
@@ -3,6 +3,7 @@
 namespace Telegram\Bot\Commands;
 
 use Illuminate\Support\Collection;
+use Mockery\Exception;
```
irazasyed, line 6: Wrong import

src/Laravel/Facades/Telegram.php
```diff
@@ -136,11 +136,6 @@
  * @method static \Telegram\Bot\Objects\WebhookInfo getWebhookInfo()
  * @method static \Telegram\Bot\Objects\Update getWebhookUpdate(bool $shouldDispatchEvents = true, ?\Psr\Http\Message\RequestInterface $request = null)
  * @method static bool removeWebhook()
- * @method static \Telegram\Bot\Objects\ForumTopic createForumTopic(array $params)
- * @method static bool editForumTopic(array $params)
- * @method static bool closeForumTopic(array $params)
- * @method static bool reopenForumTopic(array $params)
- * @method static bool deleteForumTopic(array $params)
```
irazasyed, line 143: Why are these removed?

src/Methods/Message.php
```diff
@@ -2,7 +2,6 @@
 
 namespace Telegram\Bot\Methods;
 
-use Illuminate\Support\Arr;
```
irazasyed, line 5: Why is this removed? It's in use.

irazasyed (changes requested):
Please fix the requested changes

PR #1195, opened 2026-07-15
Update sticker pack API support

e4se:
## Summary

- update sticker pack methods to the current Telegram Bot API contract
- add InputSticker support with nested multipart attach references and mixed sticker formats
- add the current sticker set management endpoints and response fields
- preserve setStickerSetThumb as a compatibility alias
- prevent InputSticker sticker values from being inferred as Sticker relations

## Testing

- 70 tests passed with 228 assertions
- added request serialization, multipart upload, object relation, and accessor regression coverage


irazasyed:
Thanks

